PR #28, opened 2020-10-09
feat: export candid

chenyan-dfinity:
The main implementation is in https://github.com/dfinity/candid/pull/119. `cdk-rs` is a simple wrapper around the candid derive. We have two ways to get the did file:
* Install the canister then call `__get_candid_interface_tmp_hack`
* Compile the canister code in WASI mode, and run wasmtime to output the did file


seniorjoinu:
Is it possible to make this work gradually?
Each procedural macros like: `init`, `query` or `update` could write to some file when they are invoked. 

I mean... for example:
* the program meets `#[init]` proc macro
* since this macros has access to function's signature, it should be able to generate a portion of candid like 
```
type A = ...;

service : (A) -> {
}
```
* then the program meets some `#[update]` proc macro and updates this file once again
```
type A = ...;
type B = ...;
type C = ...;

service : (A) -> {
  "some_update_function" : (B) -> (C);
}
```
* program continues till the end populating this file in one compilation


What do you think?

chenyan-dfinity:
The candid type is generated at runtime via `T::ty()`, so at the proc macro phase, the compiler doesn't know anything about candid type. That's why we need to compile to a WASI target to generate the did file.

neokree:
Hello @chenyan-dfinity, do you have updates on this feature?

lastmjs:
Yes any update on being able to generate the did file without having to compile once and execute code at runtime? Shouldn't static analysis of the Rust code be enough to generate a did file without any compilation?

This is a considerable headache for Azle and Kybra the TypeScript and Python CDKs

chenyan-dfinity:
> any update on being able to generate the did file without having to compile once and execute code at runtime?

No plans for this. Would it help if `dfx build` bundles the two compilation in one command?

> Shouldn't static analysis of the Rust code be enough to generate a did file without any compilation?

Theoretically yes. But you have to patch a static analyzer like `rust-analyzer` to achieve this. And you don't really gain anything. You still need to compile the code twice, one for generating the did file, one for the real wasm module.

lastmjs:
> Theoretically yes. But you have to patch a static analyzer like rust-analyzer to achieve this. And you don't really gain anything. You still need to compile the code twice, one for generating the did file, one for the real wasm module.

Can't you just use a Rust parser and use that to compile down to Candid? Or am I missing something?

lastmjs:
My real goal is for dfx to completely abstract away the need for a Candid file when writing a Rust or custom canister.

Actually today I figured out a way to do this for Azle that is pretty nice, but it would be great to avoid doing it. We take the Wasm module and stub out all of the ic0 imports, and then call a couple functions to return a pointer and the length of the pointer to the string stored on the heap after callined `__export_service`. This has made the flow a lot simpler, but still unfortunate to have to do this.

lastmjs:
> Okay, our goal is the same. You are not really oppose to getting the metadata by running an executable, which will be done in dfx build automatically.

Fantastic! We just have to do it ourselves for now using Node.js V8 Wasm engine for Azle and wasmer or wasmtime in Kybra, it will  be awesome once we can remove that code once dfx build can do this automatically!

I imagine it will be relatively simple for dfx build to do this automatically, it's been pretty easy to just run the bare module through a wasm runtime with stubbed imports.

PR #31, opened 2020-10-17
chore: remove candid crate from examples

chenyan-dfinity:
Export candid from `ic_cdk::candid`, and user projects are encouraged to use this alias for candid. This avoids problems when the candid version is different on the user project from the one used in CDK.

### DIFF
--- a/src/ic-cdk-macros/Cargo.toml
+++ b/src/ic-cdk-macros/Cargo.toml
@@ -16,7 +16,7 @@ include = ["src", "Cargo.toml", "LICENSE", "README.md"]
 proc-macro = true
 
 [dependencies]
-candid = { git = "https://github.com/dfinity/candid.git", branch = "derive-rename", version = "0.6.7", features = ["cdk"] }
+candid = "0.6.2"
 ic-cdk = { path = "../ic-cdk", version = "0.1" }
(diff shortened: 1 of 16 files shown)


src/ic-cdk-macros/Cargo.toml
```diff
@@ -16,7 +16,7 @@ include = ["src", "Cargo.toml", "LICENSE", "README.md"]
 proc-macro = true
 
 [dependencies]
-candid = "0.6.2"
+candid = { git = "https://github.com/dfinity/candid.git", branch = "derive-rename", version = "0.6.7", features = ["cdk"] }
```
hansl, line 19: What? Leftover from another branch?

src/ic-cdk/src/lib.rs
```diff
@@ -15,6 +15,7 @@ static mut DONE: bool = false;
 pub mod export {
     pub use ic_types::Principal;
 }
+pub use candid;
```
hansl, line 18: Please put this above in the `export` section.

hansl:
I'm trying to scope the re-export symbols in a `export` module (like we do for Principal). Do you mind putting it in there?

chenyan-dfinity:
Done. PTAL

hansl (approved):

PR #35, opened 2020-11-02
wip: proof of concept to create DID from types

hansl:


src/ic-cdk-macros/src/export.rs
```diff
@@ -75,6 +79,112 @@ fn get_args(method: MethodType, signature: &Signature) -> Result<Vec<(Ident, Box
...
+fn build_type(ident: Option<Ident>, typ: Type) -> IDLType {
+    match typ {
+        Type::Array(TypeArray { elem, .. }) => {
+            let elem_ty = build_type(None, *elem);
+            IDLType::VecT(Box::new(elem_ty))
+        }
+        Type::Reference(TypeReference { elem, .. }) => build_type(ident, *elem),
+        Type::Path(TypePath { path, .. }) => {
+            use candid::parser::types::PrimType;
+            match path.to_token_stream().to_string().as_str() {
```
chenyan-dfinity, line 120: A fundamental problem for working with tokens directly is that we are missing the def-use info. When you see a token `Result`, we don't know which type it refers to. Even if you require fully qualified types and matching by string (which is fragile), it still cannot handle struct/enum/tuples types, type aliases, and instantiated polymorphic types. It is also hard to match types that are "transparent" to Candid, e.g. `Box<String>`, `Rc<candid::Int>` and `Box<Vec<Box<String>>>`. Not to mention recursive types and type tokens that are generated by other macros.

Please see https://github.com/dfinity/candid/pull/119 for generating types from candid crate. The work required in CDK is minimal and outlined in https://github.com/dfinity/cdk-rs/pull/28

hansl: Yours require executing the code. While that works and isn't a lot of code for us, it's not a good workflow for developers.

Now what I'm starting to think of is something even more complex as an additional step, similar to how clippy works, for using rustc to resolve types. I don't know if we want a temporary solution (hand writing did files isn't that complicated).

chenyan-dfinity: > Yours require executing the code.

That's one option. You can also get the did file after installing the canister via `dfx call`. 

> it's not a good workflow for developers.

I think we will need a dedicated `dfx build` for Rust anyway. Inside dfx build, it doesn't make a difference whether you are running a standalone binary or the binary you just build. Thanks to incremental compilation, the second build time is also minimal, as we are only switching a flag on Rust CDK.

> similar to how clippy works, for using rustc to resolve types.

This sounds more complicated than my approach, and the benefit is simply not needing to run user's own binary?

> hand writing did files isn't that complicated

People are starting to automate this anyway, see the NNS canister: https://github.com/dfinity-lab/dfinity/tree/master/rs/nns/candid. We can and should provide something better than this.

PR #36, opened 2020-11-19
feat: Update to ic-ref-0.14

akhi3030:
Especially renaming the funds related APIs to cycles.

Fixes https://github.com/dfinity/cdk-rs/issues/38

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -119,7 +119,9 @@ pub fn call_raw(
         );
 
         ic0::call_data_append(args_raw.as_ptr() as i32, args_raw.len() as i32);
-        ic0::call_cycles_add(payment as i64);
+        if payment > 0 {
+            ic0::call_cycles_add(payment as i64);
+        }
         ic0::call_perform()
     };
 
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -201,25 +201,16 @@ pub fn reply<T: ArgumentEncoder>(reply: T) {
     }
 }
 
-/// Economics.
-///
-/// # Warning
-/// This section will be moved and breaking changes significantly before Mercury.
-/// The APIs behind it will stay the same, so deployed canisters will keep working.
-pub mod funds {
-    use super::ic0;
+pub fn msg_cycles_available() -> i64 {
+    unsafe { ic0::msg_cycles_available() }
+}
 
-    pub fn available() -> i64 {
-        unsafe { ic0::msg_cycles_available() }
-    }
+pub fn msg_cycles_refunded() -> i64 {
+    unsafe { ic0::msg_cycles_refunded() }
+}
 
-    pub fn refunded() -> i64 {
-        unsafe { ic0::msg_cycles_refunded() }
-    }
-
-    pub fn accept(max_amount: i64) -> i64 {
-        unsafe { ic0::msg_cycles_accept(max_amount) }
-    }
+pub fn msg_cycles_accept(max_amount: i64) -> i64 {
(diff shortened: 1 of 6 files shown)


src/ic-cdk/src/api/call.rs
```diff
@@ -119,12 +119,7 @@ pub fn call_raw(
         );
 
         ic0::call_data_append(args_raw.as_ptr() as i32, args_raw.len() as i32);
-
-        if payment > 0 {
-            let bytes = vec![0u8];
-            ic0::call_funds_add(bytes.as_ptr() as i32, bytes.len() as i32, payment as i64);
-        }
-
+        ic0::call_cycles_add(payment as i64);
```
nomeata, line 122: Maybe worth keeping the `if payment > 0`, to avoid pointless sytem api calls?

src/ic-cdk/src/api/ic0.rs
```diff
@@ -70,45 +70,39 @@ ic0_module! {
...
-
-    ic0.ingress_rate_bucket_size : () -> i32;                                   // Ll
-    ic0.ingress_rate_bucket_copy : (dst : i32, offset : i32, size : i32) -> (); // Ll
-    ic0.return_ingress_bucket :
-      ( bucket_src : i32,
-        bucket_size : i32,
-        cost : i64
-      ) -> ();                                                                  // Lb
-    ic0.return_ingress_limit_per_second : (limit : i64) -> ();                  // Ll
+    ic0.canister_cycle_balance : () -> i64;                                     // *
```
nomeata, line 79: Surprised to see `ingress_bucket` mentioned here (and removed by you), that has not yet been part of a released spec. Did someone copy from the `master` version of the spec, instead of a suitable release?

nomeata (commented):
LGTM, but I’ll leave approval to the actual owners of this code

src/ic-cdk/src/api/call.rs
```diff
@@ -212,33 +207,16 @@ pub fn reply<T: ArgumentEncoder>(reply: T) {
...
-                Unit::Cycle => vec![0],
-                Unit::IcpToken => vec![1],
-            }
-        }
-    }
-
-    pub fn available(unit: Unit) -> i64 {
-        let bytes = unit.to_bytes();
-        unsafe { ic0::msg_funds_available(bytes.as_ptr() as i32, bytes.len() as i32) }
+    pub fn available() -> i64 {
```
dsarlis, line 210: Nit: It might be worth naming this now `cycles_available` to make it clear what this number that it returns is. Previously, we had `available(unit)` at the call site so it was self-describing via the argument.

dsarlis (commented):
LGTM. Approval left for one of the owners.

src/ic-cdk/src/api/ic0.rs
akhi3030: That is my guess.

src/ic-cdk/src/api/call.rs
akhi3030: As this is inside the `funds` module, I thoughts it's fine.  I do not quite understand the goal of the module though.  Maybe the module should be renamed to `cycles`.  I will let the repo owners respond on this.

hansl: The module should probably be removed. It was when we had plans to have multiple types of tokens held by canisters. 

This library needs a lot of love right now. Lots of tech debt from even when I was building the wallet.

hansl (approved):
LGTM. Let me know when you think this is ready to merge.

src/ic-cdk/src/api/call.rs
akhi3030: Cool, updated.

hansl (approved):

PR #42, opened 2021-01-06
chore: set syn package's patch version

hansl:
It broke when they removed the export module from root, but we don't need it.


hansl:
This might be why https://github.com/dfinity/agent-rs/pull/88 is not building.

ericswanson-dfinity (approved):

PR #47, opened 2021-02-23
fix DFINITY Rust CDK repo url

plitzenberger:


hansl (approved):
This shouldn't break CI, LGTM.

PR #52, opened 2021-03-06
feat: expose canister certification API

roman-kashitsyn:
This change adds minimalistic certification API based on the function
that were already exported in the ic0 module.

* set_certified_data can only be called from initialization
  hooks/callbacks/update calls. It copies the data to the system for
  subsequent certification that happens asynchronously.

* data_certificate can only be called from query calls. It requests a
  certificate authenticating the previously set data from the system.


hansl (approved):
I think the Netlify doc publishing CI Action is not working for forks. I'll override this. Thanks for the API! I hope to make it more dev friendly after the launch.

PR #57, opened 2021-03-16
feat: add a static binaryen wasm-opt step that doesnt require an external bin

hansl:
Reduces further by removing some sections and applying some more optimizations. Compared with before (without the wasm-opt step) this resulted in 7% less WASM size with the wallet-rs project (around 65kiB).


p-shahi (approved):
cool

PR #59, opened 2021-03-17
feat: remove wee_alloc as the global allocation in CDK

hansl:
The CDK is now BYOA. We'll need to publish this under a new minor version.


p-shahi (approved):
We can remove wee_alloc from cargo toml too

PR #76, opened 2021-05-03
feat: add ic-certified-map package

roman-kashitsyn:
This change introduces a new package providing a merkleized map data
structure that can be used by canister authors to implement certified
queries.


hansl (approved):
LGTM we can add the documentation later.

hansl:
You can ignore the PR Title task. Apparently it doesn't work well across forks..

PR #95, opened 2021-06-03
feat: add support for inspect_message api

roman-kashitsyn:
This change

  1. Introduces a new macro (`inspect_message`) that registers
     a function that will be invoked for each ingress message.

  2. Updates the IC0 module to the latest version to get access to new
     functions.

  3. Adds 2 new api functions that can only be invoked in the context
     of a inspect_message method:

     * method_name() returns the name of the method being called.
     * accept_message() accepts the message being inspected.

See [1] for more details on how message inspection works.

I tested the new functionality by adding inspect_message to the assets
canister example, inspecting the generated Wasm and interacting with
the canister using DFX.  We really need some infra to test new CDK
features.

[1]: https://sdk.dfinity.org/docs/interface-spec/index.html#system-api-inspect-message

### DIFF
--- a/src/ic-cdk-macros/src/export.rs
+++ b/src/ic-cdk-macros/src/export.rs
@@ -20,13 +20,17 @@ enum MethodType {
     PostUpgrade,
     Update,
     Query,
+    InspectMessage,
 }
 
 impl MethodType {
-    pub fn is_lifecycle(&self) -> bool {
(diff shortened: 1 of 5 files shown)


roman-kashitsyn:
Close as this was implemented by @hpeebles in a later PR https://github.com/dfinity/cdk-rs/pull/127

PR #96, opened 2021-06-03
fix #91: differentiate between a reject and a call_new failure

### LINKED ISSUE #91: Reject codes confused with error codes returned by `ic0.call_perform`
I see this code:

```    // 0 is a special error code meaning call_simple call succeeded.
    if err_code != 0 {
        let mut state = state.borrow_mut();
        state.result = Some(Err((
            RejectionCode::from(err_code),
            "Couldn't send message".to_string(),
        )));
    }
```

This confuses synchronous error codes as returned by `invoke_call` (currenly always 0 or 1) with the [reject codes](https://sdk.dfinity.org/docs/interface-spec/index.html#reject-codes) that come with, well, a reject _message_.

Furthermore, this conflation doesn’t allow the programmer to react differently to synchronous and asynchronous errors, which is very important (e.g. you can use `api::call::msg_cycles_refunded` only in a callback).

roman-kashitsyn:
This change makes it possible to differentiate between a failure of
the call_new method and a real reject that is handled by a callback.
This is achieved by changing the error type that call() returns.

THIS IS A BREAKING CHANGE.

Fixes #91.

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -160,9 +160,18 @@ impl From<u32> for RejectionCode {
     }
 }
 
+/// Describes an outcome of an unsuccessful inter-canister call.
+pub enum CallError {
+    /// The message was accepted by the system API call, but was later
+    /// rejected either by the system or by the receiver.
+    Rejected(RejectionCode, String),
+    /// The message was rejected by the system API call.
+    SendFailed,
+}
+
 /// The result of a Call. Errors on the IC have two components; a Code and a message
 /// associated with it.
-pub type CallResult<R> = Result<R, (RejectionCode, String)>;
+pub type CallResult<R> = Result<R, CallError>;
 
 /// Internal state for the Future when sending a call.
 struct CallFutureState<R: serde::de::DeserializeOwned> {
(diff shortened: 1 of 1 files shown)


src/ic-cdk/src/api/call.rs
```diff
@@ -160,9 +160,18 @@ impl From<u32> for RejectionCode {
     }
 }
 
+/// Describes an outcome of an unsuccessful inter-canister call.
+pub enum CallError {
+    /// The message was accepted by the system API call, but was later
+    /// rejected either by the system or by the receiver.
```
nomeata, line 166:
```suggestion
    /// rejected asynchronously either by the system or by the receiver.
```

src/ic-cdk/src/api/call.rs
```diff
@@ -160,9 +160,18 @@ impl From<u32> for RejectionCode {
...
+    Rejected(RejectionCode, String),
+    /// The message was rejected by the system API call.
```
nomeata, line 168:
```suggestion
    /// The message was synchronously rejected by the system API call.
```

nomeata (commented):
Can't you make the types even more true, by letting the function rerun either a (sync) error _or_ a promise? Or is that too unwieldy?

roman-kashitsyn:
> Can't you make the types even more true, by letting the function rerun either a (sync) error _or_ a promise? Or is that too unwieldy?

It's definitely possible to change the signature of call_raw to reflect that, but I don't think it's possible to change the signature of `async fn call(...) -> CallResult<R>` because the future is implicit in the function attribute. So the CallResult will still have to be imprecise, which probably defeats the whole point...

nomeata:
Can you make a non `async` function that returns a local `async` function then? (Maybe I am thinking too functionally…)

PR #116, opened 2021-08-07
Update multiply-dependency.adoc

codebenderhq:
On the https://sdk.dfinity.org/docs/rust-guide/multiply-dependency.html site when follolwing the documentation it shows the #[update] instead of #[query] and throws an error that there is no query read function available in the rust_deps canister


adamspofford-dfinity:
Closing as the Candid file was updated to match the Rust file.

PR #122, opened 2021-08-12
fix: allow owned leaves in the certified map hash trees

### LINKED ISSUE #121: certified map: allow both references and values in HashMap::Leaf
Currently, [`HashMap::Leaf`](https://github.com/dfinity/cdk-rs/blob/64c7524b7ebb97748be1affd9ee1db1e1361d6f1/src/ic-certified-map/src/hashtree.rs#L18) only supports references.
Some clients want to be able to have a map of type `RbTree<Key, Type>` and convert values of `Type` to blobs on demand when the hash tree is constructed.
A simple solution is to store `std::borrow::Cow<'a, [u8]>` in the leaf.

roman-kashitsyn:
Currently, all the values in the hash trees produced by the certified
map must be references.  This makes it harder for the clients to
build hash trees on the fly from the contents stored in the map.

This change modifies the definition of the HashTree to allow both
references and values to be stored in the tree.

This is a BREAKING CHANGE, some client code might need to be modified
and recompiled.

Fixes #121.

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -11,7 +11,7 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: hansl/conventional-pr-title-action@v2.2.0
+      - uses: dfinity/conventional-pr-title-action@v2.2.3
         with:
           success-state: Title follows the specification.
           failure-state: Title does not follow the specification.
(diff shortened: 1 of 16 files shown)


ielashi (approved):
LGTM, thanks!

roman-kashitsyn:
@ericswanson-dfinity, could you please merge this PR? I don't think I have enough permissions for that.

PR #125, opened 2021-08-16
feat: implement proper range queries for certified map

roman-kashitsyn:
This change implements key-value range queries for the certified map
and fixes the handling of key range queries.

Before this change, the key range queries always included key
neighbors into the witness, even if the actual key bound is present in
the tree.

For example, if the tree contains keys '0'..'9', key_range('1', '8')
would produce a witness with keys '0'..'9' instead of just '1'..'8'
with everything else pruned.

This made it hard/impossible to use this map in some applications,
e.g., for registry certification.


roman-kashitsyn:
@ericswanson-dfinity, could we please merge this? I'd like to use this to improve our registry canister.

PR #134, opened 2021-09-03
refactor: remove all uses of unsafe code in ic-certified-map

roman-kashitsyn:
This change removes all uses of unsafe Rust by replacing raw pointers
to nodes with `Option<Box<Node>>`.

The behavior of the RbTree is not changed.


src/ic-certified-map/src/rbtree.rs
```diff
@@ -365,420 +335,392 @@ impl<K: 'static + AsRef<[u8]>, V: AsHashTree + 'static> RbTree<K, V> {
...
-            mut h: *mut Node<K, V>,
+        fn go<K: 'static + AsRef<[u8]>, V: AsHashTree + 'static>(
+            mut h: Box<Node<K, V>>,
             key: &[u8],
-        ) -> *mut Node<K, V> {
-            if key < (*h).key.as_ref() {
-                if !is_red((*h).left) && !is_red((*(*h).left).left) {
+        ) -> NodeRef<K, V> {
+            if key < h.key.as_ref() {
+                if !is_red(&h.left) && !is_red(&h.left.as_ref().unwrap().left) {
```
robin-kunzler, line 672: I noticed that in this method (but also in other places such as `balance`), `unwrap`s are used that assume that some invariants hold. 

E.g. this line assumes that `h.left` is_some. 

Would it make sense to make these invariants more explicit by e.g. using asserts or `expect` with a description of why the invariant holds? 

I'm aware that this also has been the case before and so this may not be something to address in this PR.

robin-kunzler (approved):
@roman-kashitsyn : Thanks for making the code safe! It's great especially from a security perspective as we don't have to worry about potential memory corruptions. With the type alias you are using I agree with you that the code reads well!

I reviewed all the changes in this PR. However I did not review the entire rb-tree implementation for correctness.

Please find two minor comments/questions below.

src/ic-certified-map/src/rbtree.rs
```diff
@@ -99,128 +101,98 @@ struct Node<K, V> {
...
+        match self.right.as_ref() {
+            None => Empty,
+            Some(r) => Pruned(r.subtree_hash),
         }
     }
 
-    fn null() -> *mut Self {
-        std::ptr::null::<Self>() as *mut Node<K, V>
+    fn null() -> NodeRef<K, V> {
+        None
```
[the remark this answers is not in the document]
roman-kashitsyn: I should probably just remove it, don't think this function adds much value.

src/ic-certified-map/src/rbtree.rs
roman-kashitsyn: I believe the invariant here is that we know that the key is in the tree, so left must be present. Added a debug_assert with a description.

The invariants required in balance are probably related to left-leaning property of the tree. I can make another pass and add more asserts in a separate PR, this will require some research.

robin-kunzler: Great, thanks!

PR #137, opened 2021-09-10
feat: Add support for 64 bit stable memory

dsarlis:
This patch adds support for the new 64 bit stable memory implementation. It's bare bones in the sense that it only provides the wrappers around the low level system api but should be enough for devs to start using the new api.


krpeacock (approved):

dsarlis:
Thanks for the review @krpeacock! What's the process to get a new release that will include this? Does it happen weekly or on an ad-hoc basis?

krpeacock:
Biweekly, generally

dsarlis:
> Biweekly, generally

@krpeacock Cool! Is it going to be this week or next? I'd like to understand the timeline so I can inform downstream users who are expecting this feature to be available.

krpeacock:
@dsarlis we will try to do one this week

PR #138, opened 2021-09-16
chore: release 0.3.2

lwshang:
Also the `ic-certified-map` crate will be releases with version `0.2.0` since the latest version on crates.io is still `0.1.0`.


krpeacock:
I'm definitely okay with stepping back to minor version locks for packages that we control, but I wanted to raise caution about whether we should do the same for external dependencies. @lwshang I will defer to your judgment if you want to go ahead

lwshang:
When I made the change, I thought it should not influence the actual version of a dependency to be used. But it is not the case. ("1.2.3" means ">=1.2.3, <2.0.0", but "1.2" means ">= 1.2.0, < 2.0.0"). The minimum limit are not the same. So I will add back the patch number back in a later commit.

krpeacock (approved):

roman-kashitsyn:
I wonder if the version number should be "0.4.0" in this case, given that there is new API.

roman-kashitsyn:
> Also the ic-certified-map crate will be releases with version 0.2.0
Could you please add me as a reviewer when you do the release?
I'd like to merge https://github.com/dfinity/cdk-rs/pull/139 and https://github.com/dfinity/cdk-rs/pull/141 before we do that.

lwshang:
@roman-kashitsyn I've published the releases to crates.io. We can make another release when your two PRs got merged.

PR #139, opened 2021-09-16
refactor: remove all uses of unsafe code in ic-certified-map

roman-kashitsyn:
This change removes all uses of unsafe Rust by replacing raw pointers
to nodes with `Option<Box<Node>>`.

The behavior of the RbTree is not changed.


roman-kashitsyn:
This is the same PR as https://github.com/dfinity/cdk-rs/pull/134, I had to create a new one because the CI has issues with forks.

robin-kunzler (approved):
@roman-kashitsyn: Thanks!

Approving again, as this is the same as #134 , see the discussion there.

PR #151, opened 2021-09-29
chore: Bump up dfx in CI

lwshang:


ericswanson-dfinity (approved):
Approved given that the chess example will be addressed in a future PR.

PR #153, opened 2021-10-06
feat: EXC-369: Add support for API cycles(u128)

AlexandraZapuc:
Add minimum support for API using Cycles(u128)


lwshang:
Close because the feat has been added in https://github.com/dfinity/cdk-rs/pull/154

PR #163, opened 2021-11-01
fix: deleted `Profile_2` to `Profile`

yasushisakai:
The `rust_profile` tutorial is a bit hard to trace without a
description of why we should:

```
type Profile = Profile_2
```

The tutorial still works without this line with refactoring `Profile_2` to `Profile`.

Please let me know if there are other intentions. (I'm very much interested to see when there is a case)

### DIFF
--- a/examples/profile/src/profile_inter_rs/lib.rs
+++ b/examples/profile/src/profile_inter_rs/lib.rs
@@ -4,21 +4,21 @@ use ic_cdk_macros::*;
 struct ProfileCanister;
 
 #[update(name = "getSelf")]
-async fn get_self() -> Box<Profile_2> {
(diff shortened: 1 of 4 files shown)


chenyan-dfinity (approved):
The did file was auto-generated from Rust types. The type name in Candid is immaterial, but for human readability, you are free to give it any name.

lwshang:
There are compile errors when I test locally.
The "test" CI wasn't run for this PR. We need fix that before merging this PR.

blueberrynotblue:
> The did file was auto-generated from Rust types.
so, you can auto-generate did file from Rust. Why not give it to us..We will still love it even though it doesn't work well.

PR #168, opened 2021-11-10
feat: a library of types for communicating with the ICP ledger

roman-kashitsyn:
This change introduces a new library, ic-ledger-types, which defines
types for communicating with the ICP ledger canister from Rust canisters:
  * `Address` and related types define methods to compute ledger
    addresses from principals and subaccounts.
  * `AccountBalanceArgs` and `TransferArgs` define structures that
    should be used for sending requests to the ledger canister.
  * MAINNET_LEDGER_CANISTER_ID defines the principal of the ledger
    canister on the mainnet.


chenyan-dfinity:
Is it better to move this inside the `ic-types` crate?

src/ic-ledger-types/src/lib.rs
```diff
@@ -0,0 +1,282 @@
+use candid::{CandidType, Principal};
+use serde::{Deserialize, Serialize};
+use sha2::Digest;
+use std::fmt;
+use std::ops::{Add, AddAssign, Sub, SubAssign};
+
+/// The subaccont that is used by default.
+pub const DEFAULT_SUBACCOUNT: Subaccount = Subaccount([0; 32]);
+
+/// The dafault fee for ledger transactions.
```
JensGroth, line 10:
```suggestion
/// The default fee for ledger transactions.
```

roman-kashitsyn:
> Is it better to move this inside the `ic-types` crate?

I tried that, but it doesn't work very well: I need to derive `CandidType` for type in this library, which would create a circular dependency.

src/ic-ledger-types/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+///       to: AccountIdentifier::new(&caller(), &DEFAULT_SUBACCOUNT),
+///       created_at_time: None,
+///     }
+///   ).await.expect("call to ledger failed").expect("transfer failed")
+/// }
+/// ```
+pub async fn transfer(
+    ledger_canister_id: Principal,
+    args: TransferArgs,
+) -> CallResult<TransferResult> {
```
MarioDfinity, line 271: We could consider flattening the errors given that `TransferError ` is specific for this method. I don't find particularly nice the double handling, e.g. the double `.expect` above, and I don't think it adds anything useful.

MarioDfinity (approved):
I left only one comment about the result of `transfer`. Overall this looks good and I've already tested it in an example.

src/ic-ledger-types/src/lib.rs
roman-kashitsyn: I'm not sure adding a new error type like this:
```rust
pub enum TransferCallError {
    CallError(CallError),
    TransferError(TransferError),
}
```
 is worth it. This type is equivalent to the current one, and there is no huge difference in the error handling:
```rust
match transfer(...) {
  Ok(Ok(block_index)) => all_is_fine(),
  Ok(Err(TransferError::...)) => handle_transfer_error(),
  Err(CallError::...) => handle_call_error(),
}
```
That doesn't look particularly bad to me.
I'd prefer to keep the interface as close to the .did file as possible.

MarioDfinity: I meant adding a variant to TransferError so that one can do

```rust
match transfer(....) {
  Ok(block_index) => all_is_fine(),
  Err(TransferError::CallError(...)) => handle_call_error(),
  Err(TransferError::...) => handle_other_errors(),
}
```

roman-kashitsyn: That seems confusing to me, because this way `TransferError` doesn't match the .did file

MarioDfinity: You are right

src/ic-ledger-types/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+documentation = "https://docs.rs/ic-ledger"
+license = "Apache-2.0"
+readme = "README.md"
+keywords = ["internet-computer", "ledger"]
+include = ["src", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+ic-cdk = "0.3.3"
```
lwshang, line 17: Should we direct this dependency with path to the same repo?
```
ic-cdk = { path = "../ic-cdk", version = "0.3" }
```

roman-kashitsyn: Done.

Just out of curiosity, what is the difference from the cargo point of view? As far as I can see, cargo automatically replaces this dependency with `ic-cdk = "^0.3"` on publishing.
I guess using a path dependency allows the developer to change both packages and release them in one go, but I'm not 100% sure.

PR #169, opened 2021-11-12
Reimplement codegen within cdk

adamspofford-dfinity:


src/ic-cdk-macros/tests/imports.rs
```diff
@@ -0,0 +1,14 @@
+// https://github.com/dtolnay/trybuild/issues/95
+use ic_cdk_macros::import;
+
+#[import(canister_id = "aaaaa-aa", candid_path = "tests/ic.did")]
+mod mgmt {
```
chenyan-dfinity, line 5: What exactly is generated here? Are you trying to add functions to the existing module, or check for subtyping of the given interface?

Can you add a test for recursive data structure? For example, what does `type List = opt record { int; List }` translates to Rust?

src/ic-cdk-codegen/src/lib.rs
```diff
@@ -0,0 +1,553 @@
...
+            if self.decls.contains_key(name) {
+                let (ident, _) = str_to_ident(name);
+                return Ok(quote!(#ident));
+            }
+        }
+        let ret = match ty {
+            IDLType::RecordT(fields) => {
+                let name = name
+                    .map(String::from)
+                    .unwrap_or_else(|| format!("_R{}", self.next_record()));
```
chenyan-dfinity, line 107: How does user know the struct name and use it in their code?

adamspofford-dfinity: The user would use their editor's autocompletion to see what types the function uses. There really isn't a better way as far as I can tell to do that - if the user's editor can't autocomplete into proc-macro generated code, that is what the build.rs version is good for.

chenyan-dfinity: Autocompletion is not always available. And all the ids can suddenly change if you add a new line in the did file.

src/ic-cdk-macros/tests/imports.rs
chenyan-dfinity: > Adding members to the module overrides existing ones with custom definitions

I think we are coming to the same conclusion: it's better to let users write down their type definitions, and the proc macro will just check for subtype. If the user didn't write a particular method or definition in the module, it gets ignored, not imported. The type name, derive traits, `Box` vs `Rc` vs lifetime are all hard to control.

> there is a function to validate a subtype, but it requires a TypeEnv forest and other things that are convenient to fetch during parsing but not afterwards

Not sure what you mean here. `add_decl` is basically your version of `TypeEnv`, why not use `TypeEnv` directly?

src/ic-cdk-codegen/src/lib.rs
adamspofford-dfinity: What naming scheme would you suggest?

src/ic-cdk-macros/tests/imports.rs
adamspofford-dfinity: If we only go off of explicit declarations then that's vastly more boilerplate that the macro is capable of generating. It already works basically perfectly as-is, with auto-imported but overridable defaults. As long as there is a default mapping for every possible candid file, and the codegen can be treated as set-it-and-forget-it, that's much less friction for the average user - that was the primary reason I started this, so I wouldn't be stuck writing boilerplate. Anything that might be sub-optimal in a particular scenario, like Nat when the user would be happy writing u128, can just be overridden, and if they have to define that themselves to begin with then AFAICT that's strictly less functionality.
> why not use `TypeEnv` directly?
Mostly because I hadn't thought of it, and I had only been looking on the methods of TypeEnv for its functionality instead of noticing the free functions in the surrounding module. I now think type-checking will be a lot easier than I thought it would a bit ago.

adamspofford-dfinity: Actually, on second thought, I think it *is* still going to be hard, because the Candid type of a Rust type is only accessible at runtime as long as it's a trait function. The maximum it would be possible to do is crash on init if the types are wrong, and even that would require changing how `#[init]` works.

chenyan-dfinity: >  I wouldn't be stuck writing boilerplate

If that's the goal, generating an initial binding from `didc` would be much easier. The user can then modify the file as they wish. Proc macro is hard to debug and understand, and the generated code is not visible to the user. Only use proc macro when the generated code is indeed transparent to the user. But in our case, it is not.

> because the Candid type of a Rust type is only accessible at runtime as long as it's a trait function.

Combined with dfx, we can export the candid type and check for subtype. See #28

src/ic-cdk-codegen/src/lib.rs
chenyan-dfinity: Extending the interface will also cause the id to change. I don't think there is a good way to generate names unless a user is involved. Proc macro generated code is hidden from the user, which makes it difficult to work with. Generating a binding from `didc`, and make that binding part of the source code is a good direction to go. The proc macro would then just check for subtype to make ensure the binding to consistent with the did file.

chenyan-dfinity:
Overall, it's a difficult problem. I suggest we write a design doc before committing to the code.

src/ic-cdk-codegen/src/lib.rs
adamspofford-dfinity: I think there are two aspects of codegen being mixed here - that it autogenerates names, and that it's in a proc macro. Whatever solution there is for names can be implemented whether it's a proc macro or a standalone code generator, and vice versa. That's why I was told to make it a separate codegen crate instead of putting the entire thing in ic_cdk_macros - this is capable of generating binding files too. Adding members to the macro-annotated module declaration and modifying the generated code in a source file are two ways of doing the same thing. I'm just trying to get it so that, regardless of which way you use the codegen, you *don't* have to write all your own declarations when the candid file contains inline things.

src/ic-cdk-macros/tests/imports.rs
adamspofford-dfinity: See other discussion for macros, but re: dfx exporting, if an external command can generate a DID file from Rust code that's perfect, but then the actual type checking seems like it'd belong in that command or some auxiliary command, as the macro never actually operates on the code it's touching, only operating on files generated on that code by an external tool. And if the macro were to call that tool, then building the crate once would require building it twice.

src/ic-cdk-macros/src/lib.rs
```diff
@@ -245,31 +245,51 @@ pub fn inspect_message(attr: TokenStream, item: TokenStream) -> TokenStream {
     handle_debug_and_errors(export::ic_inspect_message, "ic_inspect_message", attr, item)
 }
 
-/// Import another canister as a rust struct.
+/// Import another canister as a Rust module.
 ///
-/// All public interfaces defined in corresponding candid file can be accessed through the annotated struct.
+/// The attribute can be invoked two ways: `import(canister = "foo")`, which will automatically import it from a compatible canister
+/// governed by the same DFX instance ater you run `dfx build`, and `import(canister_id = "rrkah-fqaaa-aaaaa-aaaaq-cai", candid_path = "foo.did")`,
```
lwshang, line 251: typo: ater -> after

src/ic-cdk-codegen/ic.did
```diff
@@ -0,0 +1,54 @@
+type canister_id = principal;
```
lwshang, line 1: Why do we need this file in crate root?
I saw the same file at `src/ic-cdk-macros/tests/ic.did` where it is used in a test.

adamspofford-dfinity: I think I just forgot to remove it.

lwshang (approved):

chenyan-dfinity (changes requested):
Please have a proper design doc before committing anything to code. This PR is replacing a broken design with another broken design.

lwshang (approved):

PR #171, opened 2021-11-22
docs: Add `src` to mul-deps-dfx.json before `deps.did`

mashuai:


lwshang:
This fix was included in merger PR #175

PR #180, opened 2021-12-09
feat: optimizer include name section

lwshang:
A indirect dependency `hermit-abi` just made a minor release to 0.1.20 which broke our CI. 
`hermit-abi = 0.1.20` uses rust edition 2021 while our rust version in CI doesn't support it. Including Cargo.toml to enforce the version solves the problem.

P.S. `ic-cdk-optimizer` is an executable crate. We should have included Cargo.lock for reproducible compilation.


src/ic-cdk-optimizer/src/passes/binaryen.rs
```diff
@@ -27,7 +27,7 @@ impl OptimizationPass for BinaryenPass {
             Module::read(wasm).map_err(|_| String::from("Could not load module..."))?;
 
         module.optimize(&CodegenConfig {
-            debug_info: false,
+            debug_info: true, // Reserve name section for profiling use
```
chenyan-dfinity, line 30: I think `debug_info` contains more sections than name section. For example, you can run `wasm-objdump your.wasm -h` to compare the section before and after the optimizer. It will keep the following sections altogether: 
```
   Custom start=0x0008a150 end=0x000e80a9 (size=0x0005df59) ".debug_info"
   Custom start=0x000e80ac end=0x000e8290 (size=0x000001e4) ".debug_pubtypes"
   Custom start=0x000e8294 end=0x00111d12 (size=0x00029a7e) ".debug_ranges"
   Custom start=0x00111d16 end=0x001176ad (size=0x00005997) ".debug_aranges"
   Custom start=0x001176b0 end=0x00118b0a (size=0x0000145a) ".debug_abbrev"
   Custom start=0x00118b0e end=0x0015f968 (size=0x00046e5a) ".debug_line"
   Custom start=0x0015f96c end=0x001e58e3 (size=0x00085f77) ".debug_str"
   Custom start=0x001e58e7 end=0x00212826 (size=0x0002cf3f) ".debug_pubnames"
   Custom start=0x0021282a end=0x00227f76 (size=0x0001574c) "name"
   Custom start=0x00227f78 end=0x00227fc5 (size=0x0000004d) "producers"
```

lwshang: In https://docs.rs/binaryen/latest/binaryen/struct.CodegenConfig.html, it said:
```
debug_info: bool
[−]
If set, the names section is emitted.
```
Maybe, binaryen will only keep the name section. Let me also experiment.

Issue #181, opened 2021-12-24
`ic-cdk-optimizer` version support


scott-dn:
it would be great if we could support version flag return properly.
```bash
# current
$ ic-cdk-optimizer --version
ic-cdk-optimizer

# expected
$ ic-cdk-optimizer --version
ic-cdk-optimizer 0.3.3
```

PR #182, opened 2021-12-24
feat: optimizer versioning

scott-dn:
#181


scott-dn:
closed as fixed in #196

PR #184, opened 2022-01-05
feat: expose constants for the governance and cycles minting canister Ids

hpeebles:
These 2 canister Ids are needed for creating neurons and minting cycles.

I took the canister Ids from here - https://github.com/dfinity/ic/blob/master/rs/nns/canister_ids.json


ggreif (approved):
Cross-checked with https://github.com/dfinity/ic/blob/master/rs/nns/test_utils/src/itest_helpers.rs#L303

PR #208, opened 2022-02-16
fix: Remove out-of-bounds vulnerability

adamspofford-dfinity:
Fixes SDK-319.

### DIFF
--- a/src/ic-cdk/src/api/stable.rs
+++ b/src/ic-cdk/src/api/stable.rs
@@ -2,7 +2,7 @@
 //!
 //! You can check the [Internet Computer Specification](https://smartcontracts.org/docs/interface-spec/index.html#system-api-stable-memory)
 //! for a in-depth explanation of stable memory.
-use std::io;
+use std::{error, fmt, io};
 
 /// Gets current size of the stable memory.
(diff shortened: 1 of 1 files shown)


lwshang (approved):

src/ic-cdk/src/api/stable.rs
```diff
@@ -147,28 +164,42 @@ impl io::Write for StableWriter {
...
     }
 }
 
 impl StableReader {
     /// Reads data from the stable memory location specified by an offset.
     pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, StableMemoryError> {
-        stable_read(self.offset as u32, buf);
-        self.offset += buf.len();
-        Ok(buf.len())
+        let cap = (self.capacity as usize) << 16;
```
paulyoung, line 185: I have a version of these for 64-bit memory where I did something similar to prevent out-of-bounds access.

In my version, I didn't introduce a `self.capacity` field because when the memory is grown it became overly complicated to keep in sync.

For example, using the writer may result in growing the memory. In this version,  `self.capacity` would not be updated and become out of sync.

Instead, I did the equivalent of `let cap = (stable_size() as size) << 16;` here.

PR #210, opened 2022-02-17
feat: Allow configuring export macros to not reply

adamspofford-dfinity:
Adds a `reply` field to `ic_cdk_macros`'s export macros, which controls whether `ic0.msg_reply` is called by the macro. If it is not, the canister function must (and therefore can) call that function itself. 

Also adds an `Empty` type to `ic_cdk::api::call`, for annotating not-physically-returning functions with the type they semantically return, for use with the former feature.

### DIFF
--- a/docs/modules/rust-guide/examples/profile-tutorial/profile.rs
+++ b/docs/modules/rust-guide/examples/profile-tutorial/profile.rs
@@ -1,4 +1,10 @@
-use ic_cdk::export::{candid::{CandidType, Deserialize}, Principal};
+use ic_cdk::{
+    export::{
+        candid::{CandidType, Deserialize},
+        Principal,
+    },
+    call::{self, Empty},
+};
 use ic_cdk::storage;
 use ic_cdk_macros::*;
 use std::collections::BTreeMap;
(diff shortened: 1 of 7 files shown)


ericswanson-dfinity (commented):
This approach looks great.  Can we add an example and a github workflow that uses the example to test the functionality?

adamspofford-dfinity:
Sure, I'll just change the ones I was going to change in the next PR instead.

docs/modules/rust-guide/examples/profile-tutorial/profile.rs
```diff
@@ -45,22 +51,26 @@ fn update(profile: Profile) {
...
-            return Some(p);
+            profile = Some(p);
+            break;
         }
 
         for x in p.keywords.iter() {
             if x.to_lowercase() == text {
-                return Some(p);
+                profile = Some(p);
+                break;
```
ericswanson-dfinity, line 69: Does this need to break out of the outer `for` loop?

adamspofford-dfinity: Whoops!

ericswanson-dfinity (approved):

PR #230, opened 2022-03-21
feat: Add stdin param and auto-select output file

adamspofford-dfinity:
Currently, the optimizer will only select stdin if no file is specified, which takes special effort to script around, and does not support stdout. This adds `-` input and output 'paths', which explicitly read from stdin and write to stdout respectively, following unix command conentions.

The interface also requires an output parameter, even though its primary usage is to overwrite the original file. This changes that so the output, if unspecified, will default to the input.


lwshang (approved):

lwshang:
Could you update the changelog accordingly?

PR #233, opened 2022-03-23
fix: Run inter-canister calls without awaiting

adamspofford-dfinity:
Rust futures do nothing unless polled. This means that currently, call_raw will call the function whether it's awaited or not (since it is a manual future), but call/call_with_payment only actually make the call if awaited (since it is an `async fn`). This changes the implementation to have consistent behavior across the three function flavors (the caller can still disable this behavior by wrapping it in their own async block). As a bonus this no longer lifetimes the resulting future to the canister method name.


lwshang:
Does this PR introduce breaking changes?
Could you update the CHANGELOG?

adamspofford-dfinity:
This PR does not introduce breaking changes, and in fact because of the lifetime differences with and without `async fn`, programs will now compile that did not use to.

lwshang (approved):

PR #240, opened 2022-04-05
feat: Add `grow_then_write_stable_bytes` helper method

hpeebles:
# Description

Adds a new helper method, `grow_then_write_stable_bytes`, which simplifies writing bytes to stable memory by first growing the stable memory size if needed.

Also fixes a small bug where `StableWriter` was requesting more memory pages than needed (prior to this commit it would always request the total amount needed rather than only the incremental amount needed).

# How Has This Been Tested?

The `grow_then_write_stable_bytes` function was first implemented in OpenChat where we thoroughly tested it, upgrading canisters multiple times and adding lots of data between upgrades. After each upgrade we could see that the `stable_memory_size` value for the affected canister had increased on the replica dashboard.
This function is now running in the production OpenChat deployment.

I have also locally deployed a version of OpenChat which references this branch of the cdk and can again see the upgrades working correctly.

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/stable.rs
+++ b/src/ic-cdk/src/api/stable.rs
@@ -4,7 +4,9 @@
 //! for a in-depth explanation of stable memory.
 use std::{error, fmt, io};
 
-/// Gets current size of the stable memory.
+const WASM_PAGE_SIZE_IN_BYTES: u64 = 64 * 1024;
+
+/// Gets current size of the stable memory (in WASM pages).
 pub fn stable_size() -> u32 {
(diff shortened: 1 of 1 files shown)


hpeebles:
Feel free to suggest a better name... this was the best I could come up with.

hpeebles:
I'm fairly sure the test is failing because the test runner is using a really old dfx version (0.8.1).
Is there a reason it's using such an old version?
Can we update to a later version?

lwshang (approved):

PR #241, opened 2022-04-06
feat: Add a TryFrom instance for AccountIdentifier that checks CRC-32

ninegua:
# Description

We add a TryFrom instance from [u8; 32] to AccountIdentifier that checks if the bytes has a correct CRC-32 hash. 

This is a often required feature for an application that accepts account bytes from user input.

Also the ICP ledger canister itself will fail if CRC-32 is wrong. It is better to prevent wrong input in the construction. 

# How Has This Been Tested?

A simple unit test is added.

# Checklist:


lwshang:
Could you update the CHANGELOG of `ic-ledger-types`?

ninegua:
> Could you update the CHANGELOG of ic-ledger-types?

Updated. Not exactly sure if this is worth a new release, so I left out that part.

lwshang (approved):

PR #243, opened 2022-04-07
feat: Encapsulate logic to grow stable memory inside StableWriter

hpeebles:
# Description

I think this improves upon the changes made in my previous PR (#240).
Instead of exposing `grow_then_write_stable_bytes`, I think this logic should simply be encapsulated inside `StableWriter` (which is actually where it was before #240).
So this PR simply moves the logic contained in `grow_then_write_stable_bytes` back into the body of `StableWriter.write`.
This keeps the API clean by only exposing the raw ic0 methods alongside `StableWriter` and `StableReader` which orchestrate over these methods.

# How Has This Been Tested?

I have deployed canisters locally and have upgraded them a few times while added data between each upgrade.
On the replica dashboard I can see the `stable_memory_size` grow after each upgrade.

# Checklist:


hpeebles:
Closing this PR. Am working on a better solution.

hpeebles:
#245 contains a better solution.

PR #245, opened 2022-04-07
feat: Add `BufferedStableWriter` which aims to minimise system calls

hpeebles:
# Description

This adds a new stable memory writer called `BufferedStableWriter` which first buffers bytes in canister memory and only flushes those bytes to stable memory when the buffer becomes full.
Allocating additional memory and making system calls (`stable_grow` and `stable_write`) are both expensive operations, this writer minimises both.

# How Has This Been Tested?

There are unit test which cover the new `BufferedStableWriter` and the existing `StableWriter`.
I have also tested this locally and have successfully performed a few canister upgrades, adding data between each upgrade.

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/stable.rs
+++ b/src/ic-cdk/src/api/stable.rs
@@ -12,25 +12,51 @@ use std::{error, fmt, io};
 
 const WASM_PAGE_SIZE_IN_BYTES: u64 = 64 * 1024; // 64KB
 
-trait StableMemory {
+static CANISTER_STABLE_MEMORY: CanisterStableMemory = CanisterStableMemory {};
+
+/// A trait defining the stable memory API which each canister running on the IC can make use of
+pub trait StableMemory {
+    /// Gets current size of the stable memory (in WASM pages).
     fn stable_size(&self) -> u32;
+
+    /// Similar to `stable_size` but with support for 64-bit addressed memory.
     fn stable64_size(&self) -> u64;
+
+    /// Attempts to grow the stable memory by `new_pages` (added pages).
+    ///
+    /// Returns an error if it wasn't possible. Otherwise, returns the previous
+    /// size that was reserved.
+    ///
+    /// *Note*: Pages are 64KiB in WASM.
     fn stable_grow(&self, new_pages: u32) -> Result<u32, StableMemoryError>;
+
+    /// Similar to `stable_grow` but with support for 64-bit addressed memory.
     fn stable64_grow(&self, new_pages: u64) -> Result<u64, StableMemoryError>;
+
+    /// Writes data to the stable memory location specified by an offset.
+    ///
+    /// Warning - this will panic if `offset + buf.len()` exceeds the current size of stable memory.
+    /// Use `stable_grow` to request more stable memory if needed.
     fn stable_write(&self, offset: u32, buf: &[u8]);
+
+    /// Similar to `stable_write` but with support for 64-bit addressed memory.
     fn stable64_write(&self, offset: u64, buf: &[u8]);
+
+    /// Reads data from the stable memory location specified by an offset.
     fn stable_read(&self, offset: u32, buf: &mut [u8]);
+
+    /// Similar to `stable_read` but with support for 64-bit addressed memory.
     fn stable64_read(&self, offset: u64, buf: &mut [u8]);
 }
 
 /// Gets current size of the stable memory (in WASM pages).
 pub fn stable_size() -> u32 {
-    CanisterStableMemory::default().stable_size()
+    CANISTER_STABLE_MEMORY.stable_size()
 }
 
 /// Similar to `stable_size` but with support for 64-bit addressed memory.
 pub fn stable64_size() -> u64 {
-    CanisterStableMemory::default().stable64_size()
+    CANISTER_STABLE_MEMORY.stable64_size()
 }
 
 /// A possible error value when dealing with stable memory.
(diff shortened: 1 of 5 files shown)


src/ic-cdk/src/api/stable.rs
```diff
@@ -161,6 +162,142 @@ impl io::Write for StableWriter {
...
+        let memory = CanisterStableMemory::default();
+
+        Self::with_memory(buffer_size, memory)
+    }
+}
+
+impl<M: StableMemory> BufferedStableWriter<M> {
+    pub fn with_memory(buffer_size: usize, memory: M) -> BufferedStableWriter<M> {
+        BufferedStableWriter {
+            offset: 0,
```
ulan, line 199: Should `offset` be input of this function? So that the caller is aware that all bytes in the stable memory starting from `offset` are going to be overwritten?

src/ic-cdk/src/api/stable.rs
```diff
@@ -161,6 +162,142 @@ impl io::Write for StableWriter {
...
+        Ok(())
+    }
+
+    fn grow_to_capacity_bytes(
+        &mut self,
+        required_capacity_bytes: u64,
+    ) -> Result<(), StableMemoryError> {
+        let required_capacity_pages =
+            (required_capacity_bytes + WASM_PAGE_SIZE_IN_BYTES - 1) / WASM_PAGE_SIZE_IN_BYTES;
+        let current_pages = self.capacity as u64;
```
ulan, line 278: Would it make sense to check that `self.capacity` matches the current stable memory size? To protect against multiple users of the stable memory.

src/ic-cdk/src/api/stable.rs
```diff
@@ -161,6 +162,142 @@ impl io::Write for StableWriter {
     }
 }
 
+/// A writer to the stable memory which first writes data to a buffer and flushes the buffer to
+/// stable memory each time it becomes full. This reduces the number of system calls to
+/// `stable64_write` and `stable64_grow` which have relatively large overhead.
+struct BufferedStableWriter<M: StableMemory = CanisterStableMemory> {
+    /// The offset of the next write.
+    offset: u64,
```
ulan, line 170: Nit: `stable_memory_offset_in_bytes` and `stable_memory_capacity_in_pages` would improve readability.

src/ic-cdk/src/api/stable.rs
```diff
@@ -161,6 +162,142 @@ impl io::Write for StableWriter {
...
+    }
+
+    /// Attempts to grow the memory by adding new pages.
+    pub fn grow(&mut self, added_pages: u64) -> Result<(), StableMemoryError> {
+        let old_page_count = self.memory.stable64_grow(added_pages)?;
+        self.capacity = old_page_count + added_pages;
+        Ok(())
+    }
+
+    pub fn flush(&mut self) -> Result<(), StableMemoryError> {
```
ulan, line 261: Maybe mention in the comment that this is a relatively expensive operations as it going to grow the memory.
So that the user doesn't call flush after each write.

ulan (commented):
Lgtm!

src/ic-cdk/src/api/stable.rs
```diff
@@ -107,30 +134,39 @@ pub fn stable_bytes() -> Vec<u8> {
 ///
 /// Will attempt to grow the memory as it writes,
 /// and keep offsets and total capacity.
-pub struct StableWriter {
+pub struct StableWriter<M: StableMemory = CanisterStableMemory> {
```
hpeebles, line 137: This is not a breaking change. See here - https://doc.rust-lang.org/cargo/reference/semver.html#generic-new-default

src/ic-cdk/src/api/stable.rs
hpeebles: We want to minimise calls to `stable_size`, hence the need for the `capacity` field, but each time we grow stable memory we are told the latest size and set `capacity` to that new value.
If there are multiple users of stable memory then `capacity` could be incorrectly too low, but this won't cause any issues, it will just cause `stable_grow` to be called prematurely.

src/ic-cdk/src/api/stable.rs
hpeebles: No longer relevant since I'm now using `BufWriter` from `std`

src/ic-cdk/src/api/stable.rs
hpeebles: Done.

src/ic-cdk/src/api/stable.rs
hpeebles: Done.

src/ic-cdk/src/api/stable.rs
ulan: Wouldn't that cause data corruption since we would be overwriting the bytes written by another user?

hpeebles: Each canister has its own stable memory so it is up to the canister's devs to ensure that if they are using multiple writers within their canister then they don't overwrite each other's data.
Also, canisters are single threaded so I don't think there's really a use case for ever having multiple writers (maybe if you want to write at different offsets and want to avoid the call to `stable_size` each time you switch offset, but it seems unlikely anyone would do that).

The downside is that if we were to add a call to `stable_size` within `write` then each write would be 2 system calls (or more if we need to grow) rather than 1.
What would we even do in the case where the `capacity` field doesn't match `stable_size`? That would only tell us that the canister has separately called `stable_grow`, but doesn't tell us if we'd be overwriting data that we shouldn't be.

hpeebles: In fact whenever writing to stable memory devs will need to ensure they don't overwrite data which they want to keep, regardless of if the writes are happening 'concurrently' or not.

ulan: Yep, I meant this as a help for devs to ensure that they are not mixing multiple writer/serialization libraries.

ulan (approved):

hpeebles:
Given that this affects `pre_upgrade` we need to be sure that the code works properly.

We upgraded our OpenStorage canisters to make use of this earlier today.
We then did another upgrade a few hours later which will have ran the new `pre_upgrade` code.
Both of these upgrades worked fine.

So I think the unit tests + my manual local testing + the upgrades we have now run on OpenStorage provide enough proof that this code works.

PR #251, opened 2022-05-04
chore: add asset canister did interface check

roman-kashitsyn:
This change introduces a new test that checks that the certified assets
canister library satisfies the assets.did interface expected by the dfx.

The test revealed a minor discrepancy: method "delete_asset" was called
"delete_contents" for some reason. Apparently, dfx never invokes this
method.


ericswanson-dfinity (approved):
Nice!  dfx and icx-asset don't use the `delete_asset` method because the synchronization operation deletes assets using `commit_batch` with `BatchOperationKind.DeleteAsset`.

PR #256, opened 2022-05-10
feat: adds serialization-agnostic API for reading arguments and replying

chmllr:
This patch makes `arg_data_raw` public and adds `reply_raw` to provide a way to read the raw argument bytes of a method and to reply with raw bytes.

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -205,7 +205,7 @@ fn callback(state_ptr: *const InnerCell<CallFutureState<Vec<u8>>>) {
     // Make sure to un-borrow_mut the state.
     {
         state.borrow_mut().result = Some(match reject_code() {
-            RejectionCode::NoError => unsafe { Ok(arg_data_raw()) },
+            RejectionCode::NoError => Ok(arg_data_raw()),
             n => Err((n, reject_message())),
         });
     }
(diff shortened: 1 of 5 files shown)


hpeebles (approved):

hpeebles:
LGTM!

src/ic-cdk/src/api/call.rs
```diff
@@ -476,16 +477,26 @@ pub fn msg_cycles_accept128(max_amount: u128) -> u128 {
...
-    let len: usize = ic0::msg_arg_data_size() as usize;
-    let mut bytes = vec![0u8; len as usize];
-    ic0::msg_arg_data_copy(bytes.as_mut_ptr() as i32, 0, len as i32);
-    bytes
+pub fn arg_data_raw() -> Vec<u8> {
+    unsafe {
+        let len: usize = ic0::msg_arg_data_size() as usize;
+        let mut bytes = vec![0u8; len as usize];
+        ic0::msg_arg_data_copy(bytes.as_mut_ptr() as i32, 0, len as i32);
+        bytes
```
roman-kashitsyn, line 485: I usually prefer the following pattern:
```suggestion
        let mut bytes = Vec::with_capacity(len);
        ic0::msg_arg_data_copy(bytes.as_mut_ptr() as i32, 0, len as i32);
        bytes.set_len(len);
        bytes
```
The reasoning is that we don't need to initialize the vector with zeros since we're going to overwrite the buffer anyway. I never measured the difference in cycle consumption, though.
I used this approach here, for example: https://sourcegraph.com/github.com/dfinity/ic@afa6f02ac60f0c1ab0eae76d188a7ce97558e2dc/-/blob/rs/rust_canisters/dfn_core/src/stable.rs?L88-95

roman-kashitsyn (approved):

src/ic-cdk/src/api/call.rs
chmllr: Makes sense, thanks!

Issue #188, opened 2022-01-19
One-way calls


paulyoung:
@nomeata, in your post on [Zero-downtime upgrades of Internet Computer canisters](https://www.joachim-breitner.de/blog/789-Zero-downtime_upgrades_of_Internet_Computer_canisters), you said:

> On Rust, you have to use the System API directly or wait for `cdk-rs` to provide this ability (patches welcome, happy to advise)

Please can you advise? 🙂

nomeata:
Well, provide a variant of the call functions that passes `-1` for the `reply_fun` and `reject_fun` argument to `ic0.call_new`, and of course no handling/decoding of responses

aewc:
What range is the real function pointer address of `reply_fun` and `reject_fun`? At this point I discovered the benefit of setting the pointer type to i32 instead of u32. Magic, emmm

nomeata:
It’s not really a function pointer address, but it’s a WebAssembly table entry index, and these start at 0 and (so far at least) are consecutive. So 0xffffffff is a pretty good guess at an invalid one.

PR #261, opened 2022-05-13
feat: support for one-way notifications

roman-kashitsyn:
This change adds support for calls that do not require replies.

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -241,6 +241,72 @@ fn cleanup(state_ptr: *const InnerCell<CallFutureState<Vec<u8>>>) {
     }
 }
 
+fn add_payment(payment: u128) {
+    if payment == 0 {
+        return;
+    }
+    let high = (payment >> 64) as u64;
+    let low = (payment & u64::MAX as u128) as u64;
+    unsafe {
+        ic0::call_cycles_add128(high as i64, low as i64);
+    }
+}
+
+/// Sends a one-way message that invokes `method` with arguments `args` on the principal
+/// identified by `id`, ignoring the reply.
+///
+/// Returns `Ok(())` if the message was successfully enqueued, otherwise returns a reject code.
+///
+/// # Notes
+///
+///   * The caller has no way of checking whether the destination processed the notification.
+///     The system can drop the notification if the destination does not have resources to
+///     process the message (for example, if it's out of cycles or queue slots).
+///
+///   * The callee cannot tell whether the call is one-way or not.
+///     The callee must produce replies for all incoming messages.
+///
+///   * It is safe to upgrade a canister without stopping it first if it sends out *only*
+///     one-way messages.
+pub fn notify<T: ArgumentEncoder>(
+    id: Principal,
+    method: &str,
+    args: T,
+    payment: u128,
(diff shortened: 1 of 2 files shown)


domwoe:
This addresses #188

src/ic-cdk/src/api/call.rs
```diff
@@ -241,6 +241,72 @@ fn cleanup(state_ptr: *const InnerCell<CallFutureState<Vec<u8>>>) {
...
+///   * The callee cannot tell whether the call is one-way or not.
+///     The callee must produce replies for all incoming messages.
+///
+///   * It is safe to upgrade a canister without stopping it first if it sends out *only*
+///     one-way messages.
+pub fn notify<T: ArgumentEncoder>(
+    id: Principal,
+    method: &str,
+    args: T,
+    payment: u128,
```
MarioDfinity, line 275: The doc should clarify what happens to `payment` if the notification is dropped.

src/ic-cdk/src/api/call.rs
```diff
@@ -241,6 +241,72 @@ fn cleanup(state_ptr: *const InnerCell<CallFutureState<Vec<u8>>>) {
...
+///   * The caller has no way of checking whether the destination processed the notification.
+///     The system can drop the notification if the destination does not have resources to
+///     process the message (for example, if it's out of cycles or queue slots).
+///
+///   * The callee cannot tell whether the call is one-way or not.
+///     The callee must produce replies for all incoming messages.
+///
+///   * It is safe to upgrade a canister without stopping it first if it sends out *only*
+///     one-way messages.
+pub fn notify<T: ArgumentEncoder>(
```
MarioDfinity, line 271: Shouldn't this be called `notify_with_payment` to be consistent with `call_*`? Then there can be a `notify` method that sets the `payment` to 0 like `call` does for `call_raw`.

roman-kashitsyn: Maybe it's worth it to introduce a builder API for calls.

src/ic-cdk/src/api/call.rs
roman-kashitsyn: It's a tricky question: depends on implementation. With the current implementation, the payment is reimbursed. However, if we ever change the implementation to use real one-way functions, the payment will be lost.

src/ic-cdk/src/api/call.rs
adamspofford-dfinity: If you only have a few variants, builders are more code for the implementor *and* more code for the user without much benefit.

roman-kashitsyn: Renamed this function to `notify_with_payment128` and added `notify` without payment for consistency.

MarioDfinity:
LGTM.

aewc:
Looking forward to this feature, how long will it take to merge and release the version?

Actually I implemented a similar function, but using i32::MAX to reject all callbacks:
https://github.com/aewc/cdk-rs/commit/8675222b7cac6720818f9cb9c9d61e23c8d6e0e3#diff-c26c807aa13db41cbd762007fbdbda72db3f27f5b06236ab96857bf04965ffc8R377-R402

lwshang (approved):

PR #270, opened 2022-05-24
fix: switch to correct DFX environment variable

adamspofford-dfinity:
In dfinity/sdk#2032, the environment variables were unified to always be `CANISTER_CANDID_PATH_{name}`. This removes the usage of the deprecated version, `CANISTER_CANDID_{name}`.


ericswanson-dfinity (changes requested):
Please call this out in the changelog (with this change, requires dfx >= `<some version>`)

ericswanson-dfinity (approved):

PR #293, opened 2022-07-22
feat: Derive Deserialize for RejectionCode

tim-dfn:
# Description

For RejectionCode to be used, we also need to implement `Deserialize`


lwshang:
To save time, I will bump up version and make another release in this PR.

lwshang (approved):

PR #299, opened 2022-08-08
chore: improve CI

lwshang:
# Description

Make CI faster and more maintainable.

Notable changes:
- Aggregate test, fmt, lint, e2e into `ci.yml` and result in one required status `ci:required`.
- Parallelize `examples.yml` and run fmt for them.
- ic-cdk-optimizer build on three platforms; all other CI jobs run only on linux.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
...
+        uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: ${{ env.rust-version }}
+      - name: Run builds
+        run: |
+          cargo build -p ic-cdk-optimizer
+
+  test:
+    name: cargo test
+    runs-on: ubuntu-latest
```
ericswanson-dfinity, line 69: Is there any reason to run this only on ubuntu-latest, rather than `runs-on: ${{ matrix.os }}` ?

.github/workflows/examples.yml
```diff
@@ -7,47 +6,54 @@ on:
...
-            ~/.cargo/registry
-            ~/.cargo/git
-            **/target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}-1
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            examples/${{ matrix.project-name }}/target/
+          key: ${{ runner.os }}-${{ matrix.project-name }}
```
ericswanson-dfinity, line 35: What is the rationale for not including the hash of the Cargo.lock files"?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
...
+      - main
+  pull_request:
+
+env:
+  rust-version: 1.60.0
+
+jobs:
+  build:
+    name: cargo build
+    runs-on: ubuntu-latest
```
ericswanson-dfinity, line 15: Should we also run this on macos-latest and windows-latest?  I have definitely seen `cargo build` failures that only showed up on one environment.

ericswanson-dfinity (changes requested):

.github/workflows/ci.yml
lwshang: The crates checked in this `build` job only target `wasm32-unknown-unknown`. So I thought only run on ubuntu is sufficient. And ubuntu is the fastest.

.github/workflows/examples.yml
lwshang: `cdk-rs` has `Cargo.lock` in `.gitignore`. So `${{ hashFiles('**/Cargo.lock') }}` will always be empty string.
I'm considering using `${{ hashFiles('**/Cargo.toml') }}` instead. Will have a try.

.github/workflows/ci.yml
lwshang: Crates in this repo (except for `ic-cdk-optimizer`) are for `--target wasm32-unknown-unknown`. So I thought run tests on multiple platforms doesn't provide that much benefit in test coverage. I just chose the fastest platform `ubuntu`.
A trade-off between CI time and test coverage.

.github/workflows/examples.yml
smallstepman: for maximum speed, perhaps it might be worth considering to hash `cargo metadata --no-deps | jq ".packages[0].dependencies"`, that way a version bump won't cause new cache entry. 

`jq "[...].dependencies"` includes both `dependencies` and `dev-dependencies`

smallstepman: tho now I think about it, the solution may not be bullet-proof, e.g. consider what happens when you bump `rustc` version in `rust-toolchain.toml`. 
For this reason, it may be a good idea to instead include `Cargo.toml` in git tree and use that.

smallstepman: I haven't read through it yet, but you also might want to check out https://github.com/Swatinem/rust-cache

ericswanson-dfinity (approved):

.github/workflows/examples.yml
lwshang: Most crates in this repo are libraries which should better not to include Cargo.lock in version control.
The uses of `restore-keys` basically guarantees that there is always some cache to be restored. Producing new cache entries won't do much harm.
Including `rust-toolchain.toml` in the hash is a good idea. I will add it.

PR #301, opened 2022-08-10
fix: do not trap in call() if decoder fails

roman-kashitsyn:
# Description

This change maps decoding failure in the call() function to
CanisterError rejection code. This way canisters have a chance to handle
decoding failures gracefully.

# How Has This Been Tested?

The change includes an end-to-end test.

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -404,7 +404,16 @@ pub fn call<T: ArgumentEncoder, R: for<'a> ArgumentDecoder<'a>>(
     let fut = call_raw(id, method, &args_raw, 0);
     async {
         let bytes = fut.await?;
-        decode_args(&bytes).map_err(|err| trap(&format!("{:?}", err)))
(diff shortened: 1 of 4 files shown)


src/ic-cdk/src/api/call.rs
```diff
@@ -404,7 +407,16 @@ pub fn call<T: ArgumentEncoder, R: for<'a> ArgumentDecoder<'a>>(
     let fut = call_raw(id, method, &args_raw, 0);
     async {
         let bytes = fut.await?;
-        decode_args(&bytes).map_err(|err| trap(&format!("{:?}", err)))
+        decode_args(&bytes).map_err(|err| {
```
lwshang, line 410: `call_with_payment` and `call_with_payment128` below need the same fix.

roman-kashitsyn: Good catch, done!

lwshang (approved):

PR #305, opened 2022-08-22
fix(ic-certified-assets): disable etags and redirects

smallstepman:
# Description

ETags functionality hasn't been implemented in icx-proxy. [Until that happens](https://dfinity.atlassian.net/browse/BOUN-446), we're dropping this feature from the asset canister.

The support for more robust redirects will come in future dfx release (https://github.com/dfinity/sdk/pull/2414) 

# How Has This Been Tested?

ran unit tests

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] I have edited the CHANGELOG accordingly.
- [ ] I have made corresponding changes to the documentation.

PR #308, opened 2022-08-29
chore: prefix decode_etag_seq with _

lwshang:
# Description

Etags was disabled in #305. The function `decode_etag_seq` was still in the codebase and resulted in a compiler warning: "function is never used: `decode_etag_seq`".

# How Has This Been Tested?

CI

# Checklist:


src/ic-certified-assets/src/state_machine.rs
```diff
@@ -564,70 +564,6 @@ impl From<StableState> for State {
     }
 }
 
-fn decode_etag_seq(value: &str) -> Result<Vec<Hash>, String> {
```
ericswanson-dfinity, line 567: We're planning to reintroduce this functionality soon.  Can we prefix this with `_` or otherwise disable the warning for this method?  This way no one will have to hunt through git history to find this in the future.

lwshang: Yes, I prefix the function with `_`.

ericswanson-dfinity (approved):

PR #342, opened 2022-11-17
feat: Implement canister timers API

adamspofford-dfinity:
Implements SDK-828.

This PR implements the user-side functionality of the new function `ic0.global_timer_set`. Unlike other functions, it is not exposed directly as it is a single global timer, so for libraries to cooperate they must all be using the same task scheduler. 

It is additionally introduced behind a crate feature as it involves an inter-canister call to self, and if timers are not used by the canister, the canister does not need to be exported.

### DIFF
--- a/src/ic-cdk/src/timer.rs
+++ b/src/ic-cdk/src/timer.rs
@@ -1,0 +1,235 @@
+//! Provides simple timer functionality for executing a function in the future.
+
+use std::{cell::RefCell, cmp::Ordering, collections::BinaryHeap, mem, time::Duration};
+
+use futures::{stream::FuturesUnordered, StreamExt};
+use slotmap::{new_key_type, KeyData, SlotMap};
+
+// To ensure that tasks are removable seamlessly, there are two separate concepts here: tasks, for the actual function being called,
+// and timers, the scheduled execution of tasks. As this is an implementation detail, this does not affect the exported name TimerId,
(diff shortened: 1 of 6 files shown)


src/ic-cdk/src/timer.rs
```diff
@@ -0,0 +1,235 @@
...
+        tasks.get_mut(task_id).map(mem::take)
+    });
+    if let Some(mut task) = task {
+        match task {
+            Task::Once(func) => {
+                func();
+                TASKS.with(|tasks| tasks.borrow_mut().remove(task_id));
+            }
+            Task::Repeated { ref mut func, .. } => {
+                func();
```
ericswanson-dfinity, line 229: Any concerns about the effects of a panic in `func()`?

adamspofford-dfinity: Yes, well-commented ones 😛 
The inter-canister call is specifically to guard against this possibility, which will catch such a trap. No locks are held across awaits, so such a trap will roll back to a coherent state; no locks are held while the function executes, so resetting the trap point by making a further inter-canister call inside it will still reset it to a coherent state. The code that removes or reschedules the timer is placed outside the inter-canister call, so no matter what happens inside it, the scheduling will behave as intended.

ericswanson-dfinity (approved):

lwshang (approved):

PR #344, opened 2022-11-22
feat: Implement composite queries

adamspofford-dfinity:
This PR adds support for composite queries, via `#[query(composite = true)]`. Composite queries can call other query functions, but cannot be called in replicated mode.

For more details about composite queries, please check these forum threads:
* https://forum.dfinity.org/t/proposal-composite-queries/15979
* https://forum.dfinity.org/t/inter-canister-query-calls-community-consideration/6754

Implements SDK-829.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,9 +126,12 @@ jobs:
         with:
           toolchain: ${{ env.rust-version }}
           components: clippy
+      - name: Install protoc
+        run: |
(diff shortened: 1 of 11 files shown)


ericswanson-dfinity (dismissed):
Please add something to the changelog about this change/keyword.

Also, please include some detail in the PR description regarding what this is about, since not every consumer of this crate has access to jira.

PR #345, opened 2022-11-28
refactor: change from pleco to tanton

chase-manning:
# Description

Pleco is no longer maintained, it has outstanding issues and hasn't had a new contribution in two and half years.  
[Tanton](https://github.com/chase-manning/tanton) is a fork of Pleco that is backward compatable and actively maintained.  
This PR changes from using Pleco to using Tanton.

# How Has This Been Tested?

All of the Pleco tests were kept with Tanton and all are passing on `stable`, `alpha` and `nightly` versions of cargo.

The CI build should run in this PR so should test everything needed from the `cdk-rs` side.

# Checklist:

### DIFF
--- a/examples/chess/src/chess_rs/Cargo.toml
+++ b/examples/chess/src/chess_rs/Cargo.toml
@@ -15,3 +15,4 @@ ic-cdk = { path = "../../../../src/ic-cdk" }
 ic-cdk-macros = { path = "../../../../src/ic-cdk-macros" }
 serde = "1.0.111"
 tanton = "1.0.0"
+getrandom = { version = "0.2", features = ["js"] } # tanton requires this to compile on wasm target
--- a/examples/chess/src/chess_rs/Cargo.toml
+++ b/examples/chess/src/chess_rs/Cargo.toml
@@ -16,5 +16,4 @@ ic-cdk = { path = "../../../../src/ic-cdk" }
 ic-cdk-macros = { path = "../../../../src/ic-cdk-macros" }
 serde = "1.0.111"
 tanton = "1.0.0"
-imp = "0.1.0"
 getrandom = { version = "0.2", features = ["js"] }
--- a/examples/chess/src/chess_rs/Cargo.toml
+++ b/examples/chess/src/chess_rs/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name = "chess_rs"
 version = "0.1.0"
-authors = ["Hans Larsen <hans@larsen.online>"]
 edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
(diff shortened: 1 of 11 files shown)


sesi200 (commented):
LGTM, but I'll let @lwshang have the final word

examples/chess/src/chess_rs/Cargo.toml
```diff
@@ -15,4 +15,5 @@ candid = "0.8.0"
 ic-cdk = { path = "../../../../src/ic-cdk" }
 ic-cdk-macros = { path = "../../../../src/ic-cdk-macros" }
 serde = "1.0.111"
-pleco = "0.5.0"
+tanton = "1.0.0"
+getrandom = { version = "0.2", features = ["js"] }
```
chase-manning, line 19: Adding support for WebAssembly for getrandom ([Docs](https://docs.rs/getrandom/latest/getrandom/#webassembly-support))

lwshang: The canister wasm execution runtime is different from in a browser when it comes to randomness.
There is currently no synchronous randomness API exposed. The only way to get "true" randomness is calling [`raw_rand`](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-raw_rand) method of the management canister.

Adding `getrandom` here won't make a difference. I'll remove it.

lwshang (approved):

examples/chess/src/chess_rs/Cargo.toml
chase-manning: @lwshang For some reason, without `getrandom = { version = "0.2", features = ["js"] }`, this test fails:

https://github.com/dfinity/cdk-rs/actions/runs/3568798481/jobs/5998064054#step:9:85

Adding it fixes the test though...

lwshang: Yes.

IMO, this is an issue with `tanton` crate. It should have a feature flag to gate its use of `rand`.

I added it back to make compiler happy. While, I believe it will cause problem when using `tanton`'s random related API in canisters.

lwshang: Enable `js` feature of `getrandom` actually adds an import section in the wasm module which is rejected by the IC replica.

So we have to fix `tanton` crate as I said above before we can merge this PR.

lwshang: After more investigation, I believe this is something need to be fixed on `getrandom` side.
It should not forbid compilation without `js` feature enabled.

`pleco` depends on older `rand`/`rand_core` which doesn't introduce `getrandom` indirectly. That's why there is no issue on the main branch.

lwshang: Found that it is possible to register custom `getrandom` function as described [here](https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html).

I copied the code and made the example test pass.

PR #356, opened 2023-01-04
fix: handle case where a repeated task is cancelled

hpeebles:
# Description

Prior to this PR, the canister timers code would assume that a repeatable task still exists after it has been executed. But sometimes developers may only want to run a repeatable task until some end point, eg. run the repeatable task until a queue has been processed, then cancel the task.
When this happens, the code panics due to it trying to `unwrap` an `Option` of a task which no longer exists.
This PR replaces the `unwrap` with an `if let Some(...)`, preventing the code from panicking.

# How Has This Been Tested?

I have tested this locally with the scenario mentioned above where I have a queue of events which are processed by a repeated task which is cancelled once the queue is empty.

# Checklist:

- [X] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/src/ic-cdk/CHANGELOG.md
+++ b/src/ic-cdk/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+- Handle scenario where a repeated canister timer task is cancelled. (#356) 
+
 ### Refactored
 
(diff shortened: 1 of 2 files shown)


hpeebles:
Fixed by #360

dfinity-berestovskyy:
Sorry, @hpeebles I was not aware about this MR...

hpeebles:
No worries!
I didn't get round to adding tests on this yet so I'm glad you made your own PR which is in a better state than this one.

PR #357, opened 2023-01-05
fix: Refactor executor

adamspofford-dfinity:
The current executor is unsound in that it requires a pointer to be unique for safety, but shares it between the polled future and the context passed to it. Fixing this requires overhauling the executor to no longer manually manage memory lifetimes; Rc is used instead. This fixes the bug exhibited in https://github.com/hpeebles/ic-error .


src/ic-cdk/src/futures.rs
```diff
@@ -1,121 +1,135 @@
...
         }
     }
 
-    fn wake_by_ref(_: *const ()) {}
-
-    fn drop(_: *const ()) {}
+    /// # Safety
+    ///
+    /// This function should only be called by a [Waker] created by [waker].
+    unsafe fn wake_by_ref(ptr: *const ()) {
```
roelstorms, line 107: Why is it necessary to also implement wake_by_ref now?

adamspofford-dfinity: Because it's expected to exist and implementations of Future might interact with it; it's not unsafe correctness but just regular correctness.

src/ic-cdk/src/futures.rs
```diff
@@ -1,121 +1,135 @@
...
+        // SAFETY:
+        // The function's contract guarantees that the pointer is an Rc to a WakerState, and that this call borrows the data.
+        // wake has the same contract, except it takes ownership instead of borrowing. Which just requires incrementing the refcount.
+        unsafe {
+            Rc::increment_strong_count(ptr as *const WakerState);
```
roelstorms, line 112: Isn't the strong_count already increased in fn wake when it's created using from_raw?
Why should it be increased here as well?

src/ic-cdk/src/futures.rs
```diff
@@ -1,121 +1,135 @@
...
+#[cfg_attr(not(target_arch = "wasm32"), allow(unused_variables, unreachable_code))]
 pub fn spawn<F: 'static + Future<Output = ()>>(future: F) {
-    let future_ptr = Box::into_raw(Box::new(future));
-    let future_ptr_ptr: *mut *mut dyn Future<Output = ()> = Box::into_raw(Box::new(future_ptr));
-    // SAFETY: The pointer is to a Box, which satisfies the pinning requirement.
-    let mut pinned_future = unsafe { Pin::new_unchecked(&mut *future_ptr) };
-    if pinned_future
+    #[cfg(not(target_arch = "wasm32"))]
+    panic!("Cannot be run outside of wasm!"); // really, just cannot be run in a multi-threaded environment
+    let pinned_future = Box::pin(future);
```
roelstorms, line 16: Why should the future be pinned? Isn't it sufficient that it has the 'static trait?

src/ic-cdk/src/futures.rs
adamspofford-dfinity: It is incremented every time it is owned in a new location. `wake` takes ownership, `wake_by_ref` does not. You can model it as `wake` taking `Rc`, `wake_by_ref` taking `&Rc`, and `increment_strong_count` as meaning `clone`. 

`from_raw` does *not* increase the strong count; the purpose of `into_raw` and `from_raw` is to convert it to and from a raw pointer *without* dropping the handle.

roelstorms:
Do you have any tests covering this code? For example tests that track the strong_counter to see it remains correct.

The bug is about having a double free which got resolved by using RC. Do we have tests to ensure we don't have memory leaks by calling increase_strong_count on the RC?

src/ic-cdk/src/futures.rs
```diff
@@ -1,121 +1,135 @@
...
+    /// Creates a new Waker.
+    pub(crate) fn waker(state: Rc<WakerState>) -> Waker {
         // SAFETY:
-        // raw_waker has the same safety requirement on ptr as this function
-        // The functions in the vtable are passed the ptr that was passed to this function
+        // The pointer is an owning, Rc-allocated pointer to a WakerState, and therefore can be passed to raw_waker
+        // The functions in the vtable are passed said ptr
         // The functions in the vtable uphold RawWaker's contract
-        unsafe { Waker::from_raw(raw_waker(ptr)) }
+        unsafe { Waker::from_raw(raw_waker(Rc::into_raw(Rc::clone(&state)) as *const ())) }
```
roelstorms, line 133: If only the into_raw is unsafe, try isolating it in a smaller unsafe block.

roelstorms:
Why is CLEANUP and AtomicBool if we are working in a single threaded environment?

src/ic-cdk/src/futures.rs
```diff
@@ -1,121 +1,135 @@
...
 
     /// # Safety
     ///
-    /// The pointer must be a unique Box-allocated pointer to a Box-allocated pointer to a `dyn Future<Output=()>`.
-    pub unsafe fn waker(ptr: *const ()) -> Waker {
+    /// This function should only be called by a [Waker] created by [waker].
+    unsafe fn drop(ptr: *const ()) {
+        // SAFETY: The function contract guarantees that the pointer is an Rc to a WakerState, and that this call takes ownership of the data.
+        unsafe {
+            Rc::from_raw(ptr as *const WakerState);
```
roelstorms, line 123: Would it be clearer to use decrement_strong_count which seems more aligned with the purpose of this line?

roelstorms:
@robin-kunzler and I had a quick review. For added clarify, would it be possible to add a reasoning for each unsafe code block. I believe it's because of the RawWaker API that requires raw pointers but maybe you could add this somewhere as comment. This is a requirement from the [DFINITY rust coding guidelines](https://docs.dfinity.systems/dfinity/spec/meta/rust.html#_avoid_unsafe_code).

Have you investigated alternatives to this Waker which requires the use of raw pointers to avoid unsafe code in general?

As you know, unsafe code bears the risk of memory corruptions and is hard to review.

That said, the pull request significantly decreases the usage of unsafe and makes for much clearer code.

src/ic-cdk/src/futures.rs
adamspofford-dfinity: It would be possible to, but I do not think it would be clearer to; discarding the result of `from_raw` is very recognizable as calling a destructor, as that is how you do it with Box or Vec.

adamspofford-dfinity:
> Why is CLEANUP an AtomicBool if we are working in a single threaded environment?

Because it's shorter, at least until `feature(local_key_cell_methods)` stabilizes. There is no actual difference.

> Have you investigated alternatives to this Waker which requires the use of raw pointers to avoid unsafe code in general?

That's `Future`'s API. You have to implement `RawWaker` in order to construct `Waker` in order to construct `Context` in order to call `Future::poll`. There is no other way of implementing async/await.

adamspofford-dfinity:
There's no way that I have found to automate a test of this kind. Only the executor knows when a value has been dropped. Manual testing however has proved to my satisfaction that all spawned tasks are dropped, including in multiple-ownership scenarios like `join_all`.

adamspofford-dfinity:
Looking closer, there does appear to be a less fundamentally unsafe internal architecture, and I may need to fully redesign it to take it. Any reasonably implemented future should not cause a memory leak, but it's not resilient against unreasonably implemented ones.

adamspofford-dfinity:
If you say so. My concerns with the design have been *mostly* alleviated anyway - the lingering potential for memory leaks is one already present in tokio, so any futures library already knows how to avoid it.

roelstorms:
Since we only review the changes for security purposes we won't approve the entire PR. Please go through standard peer review for the required approval.

From a security perspective we saw that the unsafe code which was changed is an improvement and doesn't contain memory issues. The unsafe code which is part of this change is very simple as it only containers transformations from raw pointer to the original type and manipulating the strong count of an RC.

lwshang (approved):

PR #361, opened 2023-01-18
feat: impl From<Principal> for Subaccount

hpeebles:
# Description

This PR makes it trivial to convert a Principal into a Subaccount.
Principals are often used as Subaccounts to differentiate between funds controlled by the same Principal (eg. when creating a canister the Subaccount is the Principal of the new canister's controller).

# How Has This Been Tested?

I used this forked version to successfully top up a canister using the canister's Principal as the Subaccount.
I have also added a unit test.

# Checklist:

### DIFF
--- a/library/ic-ledger-types/src/lib.rs
+++ b/library/ic-ledger-types/src/lib.rs
@@ -123,6 +123,16 @@ impl fmt::Display for Tokens {
 )]
 pub struct Subaccount(pub [u8; 32]);
 
+impl From<Principal> for Subaccount {
+    fn from(principal: Principal) -> Self {
+        let mut subaccount = [0; 32];
+        let principal = principal.as_slice();
+        subaccount[0] = principal.len().try_into().unwrap();
+        subaccount[1..1 + principal.len()].copy_from_slice(principal);
+        Subaccount(subaccount)
+    }
+}
+
 /// AccountIdentifier is a 32-byte array.
 /// The first 4 bytes is big-endian encoding of a CRC32 checksum of the last 28 bytes.
 #[derive(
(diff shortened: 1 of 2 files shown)


library/ic-ledger-types/src/lib.rs
```diff
@@ -123,6 +123,16 @@ impl fmt::Display for Tokens {
 )]
 pub struct Subaccount(pub [u8; 32]);
 
+impl From<Principal> for Subaccount {
+    fn from(principal: Principal) -> Self {
+        let mut subaccount = [0; 32];
+        let principal = principal.as_slice();
+        subaccount[0] = principal.len().try_into().unwrap();
+        subaccount[1..1 + principal.len()].copy_from_slice(principal);
+        Subaccount(subaccount)
```
ericswanson-dfinity, line 132: How about adding a test?  There are a few at the bottom of the file.

ericswanson-dfinity (changes requested):

library/ic-ledger-types/src/lib.rs
hpeebles: Yeah I considered that but I didn't think of a suitable way to test this that isn't simply duplicating the logic in the test.
There are no existing functions exposed by `Subaccount` which could be used to verify the outcome of the `Principal` to `Subaccount` conversion.

hpeebles: I could convert a `Principal` into a `Subaccount` and then check that the bytes match what is expected.
But the only way I'd be able to get the expected bytes is by running the function itself.

hpeebles: Ok I've found a way that works and have added a unit test.

ericswanson-dfinity (approved):

hpeebles:
Who merges this? Should I be merging it?

PR #376, opened 2023-02-27
chore: Update lint settings and add missing documentation

adamspofford-dfinity:
This adds the lint settings used in `ic-cdk` to the other crates in `cdk-rs`. As one of those lints is `missing_docs`, this also adds documentation to any public symbols that were missing it, notably in `ic-certified-map` and `ic-ledger-types`.


library/ic-ledger-types/src/lib.rs
```diff
@@ -30,6 +41,7 @@ pub const MAINNET_CYCLES_MINTING_CANISTER_ID: Principal =
     CandidType, Serialize, Deserialize, Clone, Copy, Hash, Debug, PartialEq, Eq, PartialOrd, Ord,
 )]
 pub struct Timestamp {
+    /// The nanosecond count.
```
ericswanson-dfinity, line 44: See https://github.com/dfinity/ic/blob/master/rs/rosetta-api/icp_ledger/ledger.did#L8

```suggestion
    /// Number of nanoseconds from the UNIX epoch in UTC timezone.
```

library/ic-ledger-types/src/lib.rs
```diff
@@ -203,11 +217,17 @@ pub struct Memo(pub u64);
 /// Arguments for the `transfer` call.
 #[derive(CandidType, Serialize, Deserialize, Clone, Debug)]
 pub struct TransferArgs {
+    /// The memo for the transaction.
     pub memo: Memo,
+    /// The amount to be transferred.
     pub amount: Tokens,
+    /// The expected fee when transferring. Should be 0.0001 ICP.
```
ericswanson-dfinity, line 224: Suggestion: crib from https://github.com/dfinity/ic/blob/master/rs/rosetta-api/icp_ledger/ledger.did#L43 (this and other fields)

```suggestion
    /// The amount that the caller pays for the transaction.
    /// Must be 10000 e8s.
```

library/ic-ledger-types/src/lib.rs
adamspofford-dfinity: I assumed there'd be no point in repeating that from the struct docs, but :shrug:

ericswanson-dfinity (approved):

PR #383, opened 2023-03-31
feat: add ic0.is_controller to ic0 and cdk-rs

sesi200:
# Description

Introduces new system API call `ic0.is_controller`as described in this [spec change](https://github.com/dfinity/interface-spec/pull/141).

Fixes [SDK-1059](https://dfinity.atlassian.net/browse/SDK-1059)

### DIFF
--- a/src/ic-cdk/src/api/mod.rs
+++ b/src/ic-cdk/src/api/mod.rs
@@ -133,3 +133,10 @@ pub fn canister_version() -> u64 {
     // SAFETY: ic0.canister_version is always safe to call.
     unsafe { ic0::canister_version() as u64 }
 }
+
+/// Determine if a Principal is a controller of the canister.
+pub fn is_controller(principal: &Principal) -> bool {
+    let slice = principal.as_slice();
+    // SAFETY: ic0.is_controller is always safe to call.
+    unsafe { ic0::is_controller(slice.as_ptr() as i32, slice.len() as i32) as bool }
+}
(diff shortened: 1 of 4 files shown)


src/ic-cdk/src/api/mod.rs
```diff
@@ -136,6 +136,7 @@ pub fn canister_version() -> u64 {
 
 /// Determine if a Principal is a controller of the canister.
 pub fn is_controller(principal: &Principal) -> bool {
+    let slice = principal.as_slice();
     // SAFETY: ic0.is_controller is always safe to call.
```
adamspofford-dfinity, line 140: This safety comment is inaccurate - see examples of the other functions working with memory ranges. If I were to pass, for example, `0` as the first parameter, that would not be safe.

mraszyk: I think the "safety" refers to the execution contexts, e.g., query, update, callback etc. and `is_controller` is "safe" because it can be executed in all such contexts.

adamspofford-dfinity: No, it does not. It refers to safety the way Rust defines it, as in, incapable of causing UB. Any `unsafe` blocks must be accompanied by a `// SAFETY:` comment describing why the code cannot cause UB; any `unsafe` functions must have a `# Safety` header in their doc comment describing how to call them in a way that does not cause UB.

adamspofford-dfinity (approved):

PR #385, opened 2023-04-12
feat!: Implement closure-based http transform API

adamspofford-dfinity:
The existing HTTP outcall transfrom API uses `std::any::type_name` in the specific way that it says not to, and which is not guaranteed to keep working. This PR changes it to a different system where a closure is used based on the context, rather than requiring the context to be interpreted directly. An optional feature is used for the new API as it requires exporting a canister method. Manual testing was performed to ensure there were no memory leaks. 

Is there a better name for these functions?

### DIFF
--- a/src/ic-cdk/src/api/management_canister/http_request.rs
+++ b/src/ic-cdk/src/api/management_canister/http_request.rs
@@ -8,6 +8,10 @@ use candid::{
 };
 use core::hash::Hash;
 use serde::{Deserialize, Serialize};
+#[cfg(feature = "http-request")]
+use slotmap::{DefaultKey, Key, SlotMap};
+#[cfg(feature = "http-request")]
+use std::cell::RefCell;
 
 /// "transform" function of type: `func (http_request) -> (http_response) query`
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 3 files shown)


src/ic-cdk/src/api/management_canister/http_request.rs
```diff
@@ -200,6 +243,47 @@ pub async fn http_request_with_cycles(
...
+/// Make an HTTP request to a given URL and return the HTTP response, after a transformation.
+///
+/// Do not set the `transform` field of `arg`. To use a Candid function, call [`http_request_with_cycles`] instead.
+///
+/// See [IC method `http_request`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-http_request).
+///
+/// This call requires cycles payment. The required cycles is a function of the request size and max_response_bytes.
+/// Check [this page](https://internetcomputer.org/docs/current/developer-docs/production/computation-and-storage-costs) for more details.
+///
+/// If the canister is on a 13-node Application Subnet, you can call [`http_request`] instead which handles cycles cost calculation under the hood.
```
lwshang, line 255:
```suggestion
/// If the canister is on a 13-node Application Subnet, you can call [`http_request_with`] instead which handles cycles cost calculation under the hood.
```

src/ic-cdk/src/api/management_canister/http_request.rs
```diff
@@ -200,6 +243,47 @@ pub async fn http_request_with_cycles(
...
+#[cfg(any(docsrs, feature = "transform-closure"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "transform-closure")))]
+pub async fn http_request_with_cycles_with(
+    arg: CanisterHttpRequestArgument,
+    cycles: u128,
+    transform_func: impl FnOnce(HttpResponse) -> HttpResponse + 'static,
+) -> CallResult<(HttpResponse,)> {
+    assert!(arg.transform.is_none());
```
lwshang, line 263: The `assert` here might cause `trap` without giving useful information.
It seems that there is no way to guarantee that the `arg.transform.is_none()` at compile time.
I would prefer to remove this `assert`.
We clearly state this behavior in the doc comment above.

src/ic-cdk/src/api/management_canister/http_request.rs
```diff
@@ -200,6 +243,47 @@ pub async fn http_request_with_cycles(
     .await
 }
 
+/// Make an HTTP request to a given URL and return the HTTP response, after a transformation.
+///
+/// Do not set the `transform` field of `arg`. To use a Candid function, call [`http_request_with_cycles`] instead.
```
lwshang, line 248:
```suggestion
/// The `transform` field of `arg` will be ignored.. To specify a transform function by name, construct with [`TransformContext::from_name`] and call [`http_request_with_cycles`] instead.
```

src/ic-cdk/src/api/management_canister/http_request.rs
```diff
@@ -179,14 +201,35 @@ pub async fn http_request(arg: CanisterHttpRequestArgument) -> CallResult<(HttpR
     .await
 }
 
+/// Make an HTTP request to a given URL and return the HTTP response, after a transformation.
+///
+/// Do not set the `transform` field of `arg`. To use a Candid function, call [`http_request`] instead.
```
lwshang, line 206:
```suggestion
/// The `transform` field of `arg` will be ignored.. To specify a transform function by name, construct with [`TransformContext::from_name`] and call [`http_request`] instead.```

File: src/ic-cdk/src/api/management_canister/http_request.rs [hunk 5]
```diff
@@ -179,14 +201,35 @@ pub async fn http_request(arg: CanisterHttpRequestArgument) -> CallResult<(HttpR
...
+///
+/// See [IC method `http_request`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-http_request).
+///
+/// This call requires cycles payment. The required cycles is a function of the request size and max_response_bytes.
+/// This method handles the cycles cost calculation under the hood which assuming the canister is on a 13-node Application Subnet.
+/// If the canister is on a 34-node Application Subnets, you may have to compute the cost by yourself and call [`http_request_with_cycles`] instead.
```
lwshang, line 212:
```suggestion
/// If the canister is on a 34-node Application Subnets, you may have to compute the cost by yourself and call [`http_request_with_cycles_with`] instead.
```

lwshang (changes requested):

src/ic-cdk/src/api/management_canister/http_request.rs
adamspofford-dfinity: If a user specifies a transform function, it will be replaced without being called. Better to fail loudly than to fail silently imo.

adamspofford-dfinity: In other words, a trap caused by incorrectly written code will never be useful to the canister's caller - the most we can hope for is to be useful to the canister's implementor, which a panic would be.

lwshang: What will the error look like? Can the implementor easily figure out the reason of trap?

adamspofford-dfinity: It would print `arg.transform.is_none()`. Maybe I should add a more descriptive message since `arg` is not exactly eye-catching even though it's in the documentation.

lwshang (approved):

PR #391, opened 2023-04-26
lazy inter-canister calls

mraszyk:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. 

Fixes # (issue)

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -57,9 +57,11 @@ pub type CallResult<R> = Result<R, (RejectionCode, String)>;
 struct CallFutureState {
     result: Option<CallResult<Vec<u8>>>,
     waker: Option<Waker>,
+    id: Principal,
     method: String,
     arg: Vec<u8>,
-    id: Principal,
+    payment: u64,
+    payment128: u128,
 }
 
 struct CallFuture {
@@ -71,6 +73,7 @@ impl Future for CallFuture {
 
     fn poll(self: Pin<&mut Self>, context: &mut Context<'_>) -> Poll<Self::Output> {
         let self_ref = Pin::into_inner(self);
+        let state_ptr = Arc::into_raw(self_ref.state.clone());
         let mut state = self_ref.state.write().unwrap();
 
         if let Some(result) = state.result.take() {
@@ -87,8 +90,9 @@ impl Future for CallFuture {
             // ic0.call_perform is always safe to call.
             let callee = state.id.as_slice();
             let method = &state.method;
-            let args_raw = &state.arg;
-            let state_ptr = Arc::into_raw(Arc::new(&state));
+            let args = &state.arg;
+            let payment = state.payment;
+            let payment128 = state.payment128;
             let err_code = unsafe {
                 ic0::call_new(
                     callee.as_ptr() as i32,
@@ -101,8 +105,15 @@ impl Future for CallFuture {
                     state_ptr as i32,
                 );
 
-                ic0::call_data_append(args_raw.as_ptr() as i32, args_raw.len() as i32);
-                //payment_func();
+                ic0::call_data_append(args.as_ptr() as i32, args.len() as i32);
+                {
+                    if payment > 0 {
+                        // SAFETY: ic0.call_cycles_add is always safe to call.
+                        // This is called as part of the call_new lifecycle, and so will not trap.
+                        ic0::call_cycles_add(payment as i64);
+                    }
+                }
+                add_payment(payment128);
                 ic0::call_on_cleanup(cleanup as usize as i32, state_ptr as i32);
                 ic0::call_perform()
             };
@@ -274,15 +285,7 @@ pub fn call_raw(
     args_raw: &[u8],
     payment: u64,
 ) -> impl Future<Output = CallResult<Vec<u8>>> + Send + Sync {
-    call_raw_internal(id, method, args_raw, move || {
-        if payment > 0 {
-            // SAFETY: ic0.call_cycles_add is always safe to call.
-            unsafe {
-                // This is called as part of the call_new lifecycle, and so will not trap.
-                ic0::call_cycles_add(payment as i64);
-            }
-        }
-    })
+    call_raw_internal(id, method, args_raw, payment, 0)
 }
 
 /// Similar to `call128`, but without serialization.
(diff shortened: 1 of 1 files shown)


adamspofford-dfinity:
This is a reversion of an intentionally implemented feature - users who want lazy execution given eager execution can wrap in an `async` block, but users who want eager execution given lazy execution are out of luck. This is particularly relevant in the case of fire-and-forget calls.

mraszyk:
> This is a reversion of an intentionally implemented feature - users who want lazy execution given eager execution can wrap in an `async` block, but users who want eager execution given lazy execution are out of luck.

@adamspofford-dfinity I see your point. On the other hand, the example I've now included in the Motivation section results in a rather unexpected behavior and it's not clear to me if it can't be exploited for code obfuscation. Could you please take a look?

Dfinity-Bjoern:
@adamspofford-dfinity this PR is one (out of at least two) possible way of dealing with the unexpected behavior, so we may choose to resolve it in a different way. But don't you agree that this is an issue that should be resolved? The fact that `writer` may return something other than `42` or that `reader` may return with no reply seems entirely unexpected from looking at the Rust code. It basically renders typing guarantees useless.

adamspofford-dfinity:
Ah, I hadn't known this was an issue, and also looking closer, the original purpose of this behavior has been superseded by `notify`. So never mind.

src/ic-cdk/src/api/call.rs
```diff
@@ -57,6 +57,11 @@ pub type CallResult<R> = Result<R, (RejectionCode, String)>;
 struct CallFutureState {
     result: Option<CallResult<Vec<u8>>>,
     waker: Option<Waker>,
+    id: Principal,
+    method: String,
+    arg: Vec<u8>,
+    payment: u64,
+    payment128: u128,
```
adamspofford-dfinity, line 64: If we're removing the closure, you may as well remove the 64-bit code altogether. Just store a u128 here and use add_payment for both.

src/ic-cdk/src/api/call.rs
```diff
@@ -68,11 +73,60 @@ impl Future for CallFuture {
...
         } else {
+            // SAFETY:
+            // `callee`, being &[u8], is a readable sequence of bytes and therefore can be passed to ic0.call_new.
+            // `method`, being &str, is a readable sequence of bytes and therefore can be passed to ic0.call_new.
+            // `callback` is a function with signature (env : i32) -> () and therefore can be called as both reply and reject fn for ic0.call_new.
+            // `state_ptr` is a pointer created via Arc::into_raw, and can therefore be passed as the userdata for `callback`.
+            // `args`, being a &[u8], is a readable sequence of bytes and therefore can be passed to ic0.call_data_append.
+            // `cleanup` is a function with signature (env : i32) -> () and therefore can be called as a cleanup fn for ic0.call_on_cleanup.
+            // `state_ptr` is a pointer created via Arc::into_raw, and can therefore be passed as the userdata for `cleanup`.
+            // ic0.call_perform is always safe to call.
```
adamspofford-dfinity, line 90: CI is failing because the safety comments got mangled. The rule is that any *unsafe block* must be preceded by a comment describing why it's safe.

src/ic-cdk/src/api/call.rs
```diff
@@ -251,58 +297,25 @@ pub fn call_raw128(
...
+    payment: u64,
+    payment128: u128,
 ) -> impl Future<Output = CallResult<Vec<u8>>> + Send + Sync {
-    let callee = id.as_slice();
     let state = Arc::new(RwLock::new(CallFutureState {
         result: None,
         waker: None,
+        id,
+        method: method.to_string(),
+        arg: args_raw.to_vec(),
```
adamspofford-dfinity, line 315: This introduces the overhead of a copy, and the arguments may be arbitrarily large. If it can be stored borrowed, it should be; otherwise, the data should be stored in Vec form to begin with, including changing the API if necessary.

src/ic-cdk/src/api/call.rs
```diff
@@ -68,11 +73,60 @@ impl Future for CallFuture {
...
+                        }
+                    }
+                    add_payment(payment128);
+                    ic0::call_on_cleanup(cleanup as usize as i32, state_ptr as i32);
+                    ic0::call_perform()
+                };
+
+                // 0 is a special error code meaning call_simple call succeeded.
+                if err_code != 0 {
+                    state.result = Some(Err((
```
adamspofford-dfinity, line 124: This is resolving the future without reporting Poll::Ready. This would indefinitely hang the operation as it will receive no more wake-up messages.

src/ic-cdk/src/api/call.rs
mraszyk: I removed the comment, but didn't see any test failures related to the comment being present.

src/ic-cdk/src/api/call.rs
mraszyk: Indeed. I've fixed that and point out that I'm storing a (cloned) result into `state.result`, too, so that repeated calls to `poll` do not attempt to fire the call again, but rather return `Poll::Ready(result)`.

src/ic-cdk/src/api/call.rs
mraszyk: I've changed the type of `args_raw` to `Vec<u8>` so that the content can be moved. Are you ok with copying the method name? (I guess a method name is typically small and can never be larger than [20k](https://ic-interface-spec.netlify.app/#system-api-module).)

src/ic-cdk/src/api/call.rs
mraszyk: Now CI started failing so I added the comment back just before the unsafe block. The comment used to say that "ic0.call_perform is always safe to call", but I wonder what that means given that `ic0.call_perform` can only be called  in the contexts `U Ry Rt T` (same for the other ic0 calls in the unsafe block).

src/ic-cdk/src/api/call.rs
adamspofford-dfinity: Can't hurt, but for future reference, the result of calling poll again after returning Poll::Ready is allowed to do anything incoherent as long as it is not UB.

src/ic-cdk/src/api/call.rs
adamspofford-dfinity: I'd feel *slightly* better if the outer functions took `impl Into<String>` but it's not nearly as important as the argument.

src/ic-cdk/src/api/call.rs
```diff
@@ -230,79 +278,36 @@ pub fn notify_raw(
 pub fn call_raw(
     id: Principal,
     method: &str,
-    args_raw: &[u8],
+    args_raw: Vec<u8>,
```
lwshang, line 281: `arg_raw` now takes `Vec` instead of slice. Seems to be an unnecessary interface change.
I'm aware that if we keep the API, we will have to clone the slice internally which may incur some overhead.
But I still prefers to avoid API change as much as possible.
@adamspofford-dfinity  WDYT?

adamspofford-dfinity: I specifically requested this. Avoiding API change is not worth the potential overhead of a possibly multi-megabyte unnecessary copy.

adamspofford-dfinity: However, I think this API is capable of working with borrowed types. Let me do some testing.

lwshang (approved):

PR #396, opened 2023-05-11
Add borrowed arg data to lazy calls

adamspofford-dfinity:
This PR is a modification to #391 that allows argument data to be borrowed. It does so by changing our userdata pointer from an Arc to a Weak; if the caller drops the borrowed data, the Weak will fail to upgrade, preventing a dangling pointer.
@lwshang @mraszyk


mraszyk (commented):
LGTM

PR #397, opened 2023-05-12
fix: Always overwrite waker

adamspofford-dfinity:
Missed 286faae in #391 that should have been reverted. Per the Future docs, if a future needs to be polled multiple times, it should only wake the most recently provided waker.

### DIFF
--- a/src/ic-cdk/src/api/call.rs
+++ b/src/ic-cdk/src/api/call.rs
@@ -122,9 +122,8 @@ impl<T: AsRef<[u8]>> Future for CallFuture<T> {
                     state.result = Some(result.clone());
                     return Poll::Ready(result);
                 }
-
-                state.waker = Some(context.waker().clone());
             }
+            state.waker = Some(context.waker().clone());
             Poll::Pending
         }
     }


lwshang (approved):

adamspofford-dfinity:
If I understand this code correctly, the unexpected results are only due to the custom waker? If you create your own execution context for a future that is intrinsically tied to the execution context, and certainly if you then don't restore the original, it's going to cause problems. The code as written is not paying attention to Future's contract; any unexpected behavior that it exhibits is due to unexpected use of the system. The only requirement is that it not produce UB, which I do not believe it does.

mraszyk:
> If I understand this code correctly, the unexpected results are only due to the custom waker?

I could not find a way to trigger the unexpected behavior with `.await` so I used a custom waker.

> The code as written is not paying attention to Future's contract;

I suppose you mean the fact that my custom waker (unlike standard wakers) does not prevent polling multiple times in different contexts (which I couldn't achieve with `.await`)? If that's the case, then I can see why this MR makes sense.

PR #400, opened 2023-05-30
fix: Fix docs.rs build

adamspofford-dfinity:
This fixes the cfgs to no longer pretend the feature is enabled when being documented (as this is incompatible with the feature requiring optional dependencies) and instead just adds `--all-features` to the doc build to get the same effect.


lwshang:
Could you bump `ic-cdk` to 0.8.1 so that we can publish once this PR merged?

lwshang (approved):

PR #401, opened 2023-06-05
feat: set sender_canister_version and support canister_info

mraszyk:
# Description

This PR makes the following changes:

- Set caller's canister version in the field `sender_canister_version` of management canister call payloads so that the caller's canister version is recorded in the canister history.
- Add management canister types and call function for `canister_info` management canister call (`CanisterInfoRequest` and `CanisterInfoResponse`)  to support this new management canister call.

# How Has This Been Tested?

This PR adds a new test canister and integration test against `StateMachine`.

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/management_canister/main/types.rs
+++ b/src/ic-cdk/src/api/management_canister/main/types.rs
@@ -30,6 +30,16 @@ pub struct CreateCanisterArgument {
     pub settings: Option<CanisterSettings>,
 }
 
+#[derive(
+    CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Default,
+)]
+pub(crate) struct CreateCanisterArgumentExtended {
+    /// See [CanisterSettings].
+    pub settings: Option<CanisterSettings>,
+    /// sender_canister_version must be set to ic_cdk::api::canister_version()
+    pub sender_canister_version: Option<u64>,
+}
+
 /// Argument type of [update_settings](super::update_settings).
 #[derive(
     CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone,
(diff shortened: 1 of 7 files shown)


src/ic-cdk/src/api/management_canister/main/types.rs
```diff
@@ -30,6 +30,16 @@ pub struct CreateCanisterArgument {
     pub settings: Option<CanisterSettings>,
 }
 
+#[derive(
+    CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Default,
+)]
+pub(crate) struct CreateCanisterArgumentExtended {
```
chenyan-dfinity, line 36: Why not adding the new option field to the original struct? It should be a backward compatible change.

mraszyk: > It should be a backward compatible change.

I don't think so as Rust does not allow default values for struct fields. This means that clients of the library would need to add this field to their code (either assigning `None` or `Some(ic_cdk::api::canister_version())` to the field). This PR saves the clients from that hassle.

chenyan-dfinity: Or we could bump the major version and let the users to update their code, which is a common practice in Rust. If we add a new struct for each API change, it's also confusing to the users which struct to use.

src/ic-cdk/src/api/management_canister/main/types.rs
```diff
@@ -135,3 +183,131 @@ pub struct CanisterStatusResponse {
...
+}
+
+/// Provides details on the respective canister change.
+#[derive(
+    CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone,
+)]
+pub enum CanisterChangeDetails {
+    /// See [CanisterCreationRecord].
+    #[serde(rename = "creation")]
+    CanisterCreation(CanisterCreationRecord),
```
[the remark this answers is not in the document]
mraszyk: I've removed `Canister` from `CanisterChangeOrigin` and `CanisterChangeDetails` variant types, but
```
/// Provides details on the respective canister change.
#[derive(
    CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone,
)]
#[serde(rename_all = "snake_case")]
pub enum CanisterChangeDetails {
    /// See [CreationRecord].
    Creation(CreationRecord),
    /// Uninstalling canister's module.
    CodeUninstall,
    /// See [CodeDeploymentRecord].
    CodeDeployment(CodeDeploymentRecord),
    /// See [ControllersChangeRecord].
    ControllersChange(ControllersChangeRecord),
}
```
makes the test `test_canister_info` fail.

adamspofford-dfinity (approved):

PR #404, opened 2023-06-20
fix: no hard-coded fees for management canister calls

lwshang:
SDK-1142

# Description

The management canister API takes explicit `cycles` argument if the call requires payment. No more hard-coded fees.

# How Has This Been Tested?

Example and e2e test

# Checklist:


src/ic-cdk/src/api/management_canister/http_request/mod.rs
```diff
@@ -0,0 +1,100 @@
+//! Canister HTTP request.
```
ericswanson-dfinity, line 1: A suggestion for future work when moving a file like this.  Open a separate, preparatory PR that renames (as in this case) `request.rs -> request/mod.rs`.  This will make it easier / possible to review the differences.  As is, it's not reasonable for a reviewer to review the changes: they must review the whole file, or manually diff the files.

ericswanson-dfinity (approved):

PR #409, opened 2023-07-04
chore: Update example http_request_required_cycles

AlexandraZapuc:
This PR update the function `http_request_required_cycles` from `examples` to use the new proposed formula specified [here](https://forum.dfinity.org/t/a-new-price-function-for-https-outcalls/20838).

### DIFF
--- a/examples/management_canister/src/caller/lib.rs
+++ b/examples/management_canister/src/caller/lib.rs
@@ -88,7 +88,8 @@ mod http_request {
         };
         let arg_raw = ic_cdk::export::candid::utils::encode_args((arg,))
             .expect("Failed to encode arguments.");
-        (3_000_000u128 + 60_000u128 * 13
+        (3_000_000u128
+            + 60_000u128 * 13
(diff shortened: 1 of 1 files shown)


examples/management_canister/src/caller/lib.rs
```diff
@@ -88,9 +88,10 @@ mod http_request {
         };
         let arg_raw = ic_cdk::export::candid::utils::encode_args((arg,))
             .expect("Failed to encode arguments.");
-        400_000_000u128 / 13
-            + 100_000u128 / 13
-                * (arg_raw.len() as u128 + "http_request".len() as u128 + max_response_bytes)
+        (3_000_000u128 + 60_000u128 * 13
```
AlexandraZapuc, line 91: I have only found this place that needs to be changed because of the new proposed formula for calculating the cost of the http outcalls. 

Because it's a small change in the examples folder, I did not update  the changelog file as it does not seem necessary.

PR #424, opened 2023-09-06
feat: sdk 1227

lwshang:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. 

Fixes # (issue)

# How Has This Been Tested?

# Checklist:


chenyan-dfinity:
Interesting idea! There are two approaches to solve the double compilation problem with different pros and cons:

1) As you proposed here, we can provide the `ic` imports in the wasmtime runtime. 
2) We can build a Wasm transformation that takes out the `ic` imports, so that it can run with regular wasmtime.

For 1), we need a special wasmtime (`candid_extractor` in this case), and the output candid function will be in the deployed canisters.
For 2), we need a regular wasmtime, which can also be used for Motoko unit testing as well. The transformation can be useful in other cases, e.g., when some crates accidentally import some non-ic modules, we can help remove them.

src/ic-cdk-timers/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [0.54.0] - 2023-09-18
```
adamspofford-dfinity, line 9:
```suggestion
## [0.5.0] - 2023-09-18
```

adamspofford-dfinity (approved):

PR #427, opened 2023-09-19
chore: CI for release candid-extractor

lwshang:
# Description

Add a workflow to release `candid-extractor` that support `cargo binstall`.

This [release](https://github.com/dfinity/cdk-rs/releases/tag/candid-extractor-v0.1.1) is a successful run of the new workflow.

# How Has This Been Tested?

Check the release above.

`cargo binstall candid-extractor` which should directly download the prebuilt binary without compilation.

# Checklist:


.github/workflows/release-candid-extractor.yml
```diff
@@ -0,0 +1,61 @@
...
+  rust-version: 1.65.0
+
+jobs:
+  release:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
```
ericswanson-dfinity, line 17:
```suggestion
          - os: ubuntu-20.04
```

.github/workflows/release-candid-extractor.yml
```diff
@@ -0,0 +1,61 @@
...
+            binstall_pkg: candid-extractor-x86_64-unknown-linux-gnu.tar.gz
+          - os: macos-latest
```
ericswanson-dfinity, line 19:
```suggestion
          - os: macos-12
```

ericswanson-dfinity (changes requested):
I think it would be better to build with releases on specific versioned runners, rather than the `-latest` variants.

For one, when GitHub updates these, they typically roll out changes gradually.  So for example, when macos-latest was transitioning from macos-11 to macos-12, maybe 10% of runs specifying macos-latest would use macos-12 runners, while the remainder used macos-11.

For ubuntu, I've suggested ubuntu-20.04 rather than ubuntu-22.04 to avoid linker errors with glibc.  This is why we build dfx with ubuntu-20.04 as well.

.github/workflows/release-candid-extractor.yml
```diff
@@ -0,0 +1,61 @@
...
+            ${{ runner.os }}-
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: ${{ env.rust-version }}
+
+      - name: Build
+        run: |
+          cargo build -p candid-extractor --release
```
ericswanson-dfinity, line 47:
```suggestion
          cargo build -p candid-extractor --release --locked
```

ericswanson-dfinity (approved):

PR #451, opened 2023-12-12
feat: "hidden" attribute to exclude exporting method in export_candid!()

chenyan-dfinity:
# Description

Add `#[query(no_export)]`/`#[update(no_export)]` attribute to exclude exporting certain endpoints in did file.

# How Has This Been Tested?

# Checklist:


lwshang:
The syntax is actually: `#[query(no_export = true)]` `#[update(no_export = true)]`.

I considered removing the `no_` prefix, i.e. `#[query(export = false)]`. If so , the `export` option will default to true which is not well supported by `#[serde(default)]`. So I keep the original name `no_export` which defaults to false.

chenyan-dfinity:
How about `hidden` instead of `no_export`?

lwshang (approved):

PR #461, opened 2024-02-01
feat: add management canister methods for interacting with the chunk store

hpeebles:
# Description

Adds `upload_chunk`, `stored_chunks`, `clear_chunk_store` and `install_chunked_code` to the management canister API.

# How Has This Been Tested?

I have tested this locally by uploading a wasm comprised of 3 chunks, the I queried `stored_chunks` which returned the chunks, then I called `install_chunked_code` which completed successfully, then I called `clear_chunk_store`, then I queried `stored_chunks` again and this time it was empty.

# Checklist:

### DIFF
--- a/src/ic-cdk/src/api/management_canister/main/types.rs
+++ b/src/ic-cdk/src/api/management_canister/main/types.rs
@@ -85,6 +85,7 @@ pub struct UploadChunkArgument {
 )]
 pub struct ChunkHash {
     /// The hash of an uploaded chunk
+    #[serde(with = "serde_bytes")]
     pub hash: Vec<u8>,
 }
 
--- a/src/ic-cdk/src/api/management_canister/main/types.rs
+++ b/src/ic-cdk/src/api/management_canister/main/types.rs
@@ -68,7 +68,9 @@ pub(crate) struct UpdateSettingsArgumentExtended {
 }
 
 /// Argument type of [update_chunk](super::update_chunk).
-#[derive(CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+#[derive(
+    CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone,
+)]
 pub struct UploadChunkArgument {
     /// The canister whose chunk store the chunk will be uploaded to
     pub canister_id: CanisterId,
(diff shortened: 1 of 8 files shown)


src/ic-cdk/src/api/management_canister/main/types.rs
```diff
@@ -117,20 +188,63 @@ pub(crate) struct InstallCodeArgumentExtended {
...
+#[derive(
+    CandidType, Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone,
+)]
+pub struct InstallChunkedCodeArgument {
+    /// See [CanisterInstallMode].
+    pub mode: CanisterInstallMode,
+    /// Principal of the canister being installed
+    pub target_canister: CanisterId,
+    /// The canister in whose chunk storage the chunks are stored (defaults to target_canister if not specified)
+    pub storage_canister: Option<CanisterId>,
```
hpeebles, line 201: The field is called `store_canister`. See here - https://github.com/dfinity/ic/blob/master/rs/types/management_canister_types/src/lib.rs#L2462

hpeebles: Hmm I see that its called `storage_canister` in the spec... one of them needs to change

src/ic-cdk/src/api/management_canister/main/types.rs
```diff
@@ -198,9 +198,9 @@ pub struct InstallChunkedCodeArgument {
     /// Principal of the canister being installed
     pub target_canister: CanisterId,
     /// The canister in whose chunk storage the chunks are stored (defaults to target_canister if not specified)
-    pub storage_canister: Option<CanisterId>,
+    pub store_canister: Option<CanisterId>,
     /// The list of chunks that make up the canister wasm
-    pub chunk_hashes_list: Vec<ChunkHash>,
+    pub chunk_hashes_list: Vec<Vec<u8>>,
```
hpeebles, line 203: This will be serialized inefficiently when using serde (it'll be fine when using candid, I think?).
We should either use `ByteBuf` or manually implement serialize/deserialize for this field.

hpeebles: Yeah ideally we'd expose `Vec<u8>` rather than `ByteBuf`, but when its `Vec<Vec<u8>>` we can't use the `#[serde(with = "serde_bytes")]` trick. I think the only way to avoid exposing `ByteBuf` would be to manually implement serialize and deserialize.

lwshang:
@hpeebles

For the inconsistency between the interface-spec and the replica implementation, the conclusion is that the replica will be modified to align with the spec. So we have to wait for that before we can complete this PR.

And since it will be `pub chunk_hashes_list: Vec<ChunkHash>` in the end, we don't have to implement the custom serde methods now.

Turning this PR back to draft now.

lwshang (approved):

PR #464, opened 2024-02-21
fix: use standard canister id environment variables

ericswanson-dfinity:
# Description

Look up the `CANISTER_ID_<canister_name_uppercase>` and `CANISTER_CANDID_PATH_<canister_name_uppercase>` 
  environment variables to get the canister ID for a canister name. Previously the case of the canister
  name matched the case of the canister, typically lowercase.

Needed for https://dfinity.atlassian.net/browse/SDK-1379

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/src/ic-cdk-bindgen/CHANGELOG.md
+++ b/src/ic-cdk-bindgen/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+- Look up the `CANISTER_ID_<canister_name_uppercase>` and `CANISTER_CANDID_PATH_<canister_name_uppercase>` 
+  environment variables to get the canister ID for a canister name. Previously the case of the canister
+  name matched the case of the canister, typically lowercase.
+
 ## [0.1.2] - 2023-11-23
 
 - Change `candid` dependency to the new `candid_parser` library.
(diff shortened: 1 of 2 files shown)


src/ic-cdk-bindgen/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+- Breaking change: Look up the `CANISTER_ID_<canister_name_uppercase>` and
```
lwshang, line 9: IIUC, this doesn't have to be a breaking change.
Before this PR, `ic-cdk-codegen` works with existing `dfx` but may stop working with future `dfx` which removes the legacy env-var.
After this PR, `ic-cdk-codegen` should work with existing and future `dfx` versions.
From a user perspective, the fix doesn't break.

ericswanson-dfinity: That's true.  The breaking behavior I was thinking of would be for someone building with the CDK but not using dfx, if they didn't set both forms of environment variable.

ericswanson-dfinity: @lwshang PTAL

lwshang: I would like to make this PR non-breaking in any sense. That we make `ic-cdk-bindgen` look for the uppercase env vars firs, if not exist then look for the legacy ones .

And simply uppercasing the canister name is not accurate. We also need to replace hyphens with underscores.

I'd like to take over the following work. WDYT?

lwshang:
Implemented in #467

PR #465, opened 2024-02-25
feat!: config canister endpoint decoding quota

chenyan-dfinity:
# Description

Follow up on https://github.com/dfinity/candid/pull/524

# How Has This Been Tested?

# Checklist:


src/ic-cdk-timers/src/lib.rs
```diff
@@ -265,7 +265,12 @@ extern "C" fn timer_executor() {
     if ic_cdk::api::caller() != ic_cdk::api::id() {
         ic_cdk::trap("This function is internal to ic-cdk and should not be called externally.");
     }
-    let (task_id,) = ic_cdk::api::call::arg_data();
+    let config = ic_cdk::api::call::ArgDecoderConfig {
+        decoding_quota: Some(10_000),
+        skipping_quota: Some(100),
```
mraszyk, line 270: Where do these numbers come from?

src/ic-cdk/src/api/call.rs
```diff
@@ -642,14 +651,49 @@ pub fn reply_raw(buf: &[u8]) {
...
-    match decode_args(&bytes) {
+    let mut config = DecoderConfig::new();
+    if let Some(n) = arg_config.decoding_quota {
+        config.set_decoding_quota(n);
+    }
+    if let Some(n) = arg_config.skipping_quota {
+        config.set_skipping_quota(n);
+    }
+    if arg_config.debug {
+        config.set_full_error_message(true);
```
mraszyk, line 678: Does this refer to decoding costs or also full types (the latter could be too much output also in the debug mode)?

src/ic-cdk/src/api/call.rs
```diff
@@ -355,6 +358,12 @@ fn decoder_error_to_reject<T>(err: candid::error::Error) -> (RejectionCode, Stri
     )
 }
 
+fn decode_args<T: for<'a> ArgumentDecoder<'a>>(bytes: &[u8]) -> candid::Result<T> {
+    let mut config = DecoderConfig::new();
+    config.set_skipping_quota(100_000); // should we enforce this or only for untrusted calls?
```
mraszyk, line 363: I think it makes sense to have it. I wouldn't expect that an API evolves in a way that it used to expect a lot of data and now it doesn't.

src/ic-cdk-timers/src/lib.rs
chenyan-dfinity: It's an arbitrary number, but this method only returns a nat, so a small limit on both would be good enough.

src/ic-cdk/src/api/call.rs
chenyan-dfinity: Debug mode is supposed to be run in local replica. It not only outputs the full types, it also outputs the decoder state in case of error. Decoding cost is only shown when there is no decoding error.

src/ic-cdk/src/api/call.rs
chenyan-dfinity: Added `call_with_config` function instead.

src/ic-cdk/src/api/management_canister/http_request/mod.rs
```diff
@@ -39,14 +39,14 @@ thread_local! {
...
-        call::{arg_data, reply},
+        call::{arg_data, reply, ArgDecoderConfig},
         caller,
     };
     if caller() != Principal::management_canister() {
         crate::trap("This function is internal to ic-cdk and should not be called externally.");
     }
     crate::setup();
-    let (args,): (TransformArgs,) = arg_data();
+    let (args,): (TransformArgs,) = arg_data(ArgDecoderConfig::default());
```
ericswanson-dfinity, line 49: Is there anything a developer needs to know about the use of this default here?

src/ic-cdk/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+    - `decoding_quota` limits the total amount of work the deserializer can perform. See [docs on the Candid library](https://docs.rs/candid/latest/candid/de/struct.DecoderConfig.html#method.set_decoding_quota) to understand the cost model.
+    - `debug = true` prints the instruction count and the decoding/skipping cost to the replica log, after a successful deserialization. The decoding/skipping cost is logged only when you have already set a quota in the attributes. The debug mode is useful to determine the right quotas above. Developers can send a few large payloads to the debugging endpoint and know the actual decoding cost.
+  * When making inter-canister calls, we have a new function `call_with_config` to config the same decoding quotas described above. It's strongly recommended to use `call_with_config` when calling third-party untrusted canisters.
```
ericswanson-dfinity, line 17: The PR title says "feat!", indicating that this is a breaking change. Is there some change that developers will need to make, that we can mention here?  The description above says "Allow setting ..." which implies it's optional to make this change.

chenyan-dfinity: The only breaking change is that `ic_cdk::api::call::arg_data` now takes `ArgDecoderConfig` instead of no argument. It's only a breaking change when people are using `arg_data` directly. I expect most users would use `arg_data` via `#[update/query]`. In that case, it's not a breaking change.

src/ic-cdk/src/api/management_canister/http_request/mod.rs
ericswanson-dfinity: Isn't the TransformArgs arbitrarily large?  While we trust the management canister, the response and the context don't come from directly from it, right?

```
pub struct TransformArgs {
    /// Raw response from remote service, to be transformed
    pub response: HttpResponse,

    /// Context for response transformation
    #[serde(with = "serde_bytes")]
    pub context: Vec<u8>,
}
```

chenyan-dfinity: Yes, we only set the skipping quota here. The decoding of expected types, i.e., HttpResponse and blob can be arbitrarily large, and doesn't count towards the skipping quota. Same as the default setting with canister endpoint, we don't set a limit for what's being requested (expected type) by the code. We only want to reject malicious messages which contain too many unneeded information.

ericswanson-dfinity (approved):

PR #480, opened 2024-04-11
refactor!: move Rust code generation logic from candid_parser.

lwshang:
SDK-1589

# Description

We are going to make `ic-cdk-bindgen` support more use cases, e.g. generating types only, generating bindings for canister implementation.

The first step is to move the Rust code generation logic from the candid_parser crate so that we can have complete control of Rust codegen.

The added `code_generator.rs` file was the [rust/candid_parser/src/bindings/rust.rs](https://github.com/dfinity/candid/blob/ae4d0f794a808b88f8b83da7eb96ca87b3d751d4/rust/candid_parser/src/bindings/rust.rs) in the candid crate.

The only difference:

```diff
diff cdk-rs/src/ic-cdk-bindgen/src/code_generator.rs candid/rust/candid_parser/src/bindings/rust.rs
0a1
> use super::analysis::{chase_actor, infer_rec};
3d3
< use candid_parser::bindings::analysis::{chase_actor, infer_rec};
```

# How Has This Been Tested?

Example tests

# Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -892,7 +892,10 @@ dependencies = [
 name = "ic-cdk-bindgen"
 version = "0.1.3"
 dependencies = [
+ "candid",
  "candid_parser",
+ "convert_case",
+ "pretty",
 ]
 
 [[package]]
--- a/src/ic-cdk-bindgen/CHANGELOG.md
+++ b/src/ic-cdk-bindgen/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Refactor!: move Rust codegen logic from candid_parser. (#480)
+- Refactor!: move Rust code generation logic from candid_parser. (#480)
 
 ### Fixed
 
--- a/src/ic-cdk-bindgen/CHANGELOG.md
+++ b/src/ic-cdk-bindgen/CHANGELOG.md
@@ -6,23 +6,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
-- Re-generate bindings if the canister ids changed (e.g. when switching networks) or when the path to the candid file of a dependency changed.
+### Changed
+
+- Refactor!: move Rust codegen logic from candid_parser. (#480)
+
+### Fixed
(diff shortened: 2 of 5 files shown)


chenyan-dfinity:
How do you make sure the bindgen stays in sync with `candid_parser::bindings::rust`? `candid_parser` is currently going through a major revamp, likely the Rust binding module will look completely different soon.

lwshang:
> How do you make sure the bindgen stays in sync with candid_parser::bindings::rust?

There is no need to keep sync. This PR is just setting a starting point for my following rewrite of `ic-cdk-bindgen` crate. I will soon make it very different from `candid_parser::bindings::rust`. 

---

Currently, we keep the language bindings in `candid_parser` crate in the candid repo. While some of the language codegen logic was for some particular downstream projects, e.g. cdk-rs. agent-rs, agent-js, etc. This introduced a circular dependency which makes it really hard to maintain both sides.

There were a few cases illustrated why it is not a good idea to have such circular dependency:

### Case 1

When we needed to change the generated Rust code in `ic-cdk-bindgen`. We had to go through following steps:
1. Open a PR in the candid repo which change the codegen logic.
2. Open a `cdk-rs` PR which updated the `candid` dependency (with a temporary git commit ref) to see if the change is as expected. 
3. Merge the `candid` PR and make a new release of `candid_parser`.
4. Change the `cdk-rs` PR to use the new release of `candid_parser` and merge the PR.

### Case 2

When there was a change in the `candid_parser` crate that accidentally altered the Rust codegen, a patch release was made. However, this change was blindly passed to the `ic-cdk-bindgen`, causing it to break. To fix the issue, we had to go through the cumbersome steps in case 1.

---

I'm envisioning a more clear separation of responsibilities which will make the projects more cohesive and reduce coupling.
- The candid repo (`candid_parser`, `didc bind`) can still keep some language codegen functionality.
  - But such codegen should be usage agnostic, hence no circular dependency.
- `ic-cdk-bindgen` is in charge of bindgen from Candid to Rust **only for the usage inside Rust CDK**.
  - If we want to change the generated Rust code, we will only need one cdk-rs PR and have proper CI check there.

chenyan-dfinity (approved):
resolved offline

adamspofford-dfinity (approved):

PR #520, opened 2024-09-25
feat(ic-ledger-types): Add AccountIdentifier.as_bytes()

### LINKED ISSUE #519: feature request: Get AccountIdentifier as bytes
# Motivation
I need an account identifier as bytes.

Given the latest published API ([v0.13.0](https://docs.rs/ic-ledger-types/latest/ic_ledger_types/struct.AccountIdentifier.html)) it appears that the only way of getting the bytes is with `account_identifier.to_hex()` and then to parse the hex.  The bytes are in `account_identifier.0` but this is private.

# Requested changes
Add a method that provides the bytes.  Something like:
```
pub fn as_bytes(&self) -> &[u8;32] {
  self.0
}
```
I believe that the design of AccountID is committed to a 32 byte array, so it seems reasonable to include that in the signature rather than making a `fn to_slice(&self) -> &[u8]` and leaving it up to every caller to check whether the slice is indeed 32 bytes.

If approved in principle but the maintainers are busy I would be happy to make the PR.  It seems simple enough :slightly_smiling_face:

bitdivine:
# Description
It is currently not possible to obtain an `AccountIdentifier` as bytes.  It is only possible to obtain hex and then convert the hex back to bytes.  This is resolved by adding an `.as_bytes()` method that returns a reference to the inner bytes.

Fixes #519

# How Has This Been Tested?
A unit test is included that verifies that parsing bytes and returning them as bytes yields the original data.

# Checklist:

  - Rustdoc is included on the method.

### DIFF
--- a/library/ic-ledger-types/src/lib.rs
+++ b/library/ic-ledger-types/src/lib.rs
@@ -218,6 +218,11 @@ impl AccountIdentifier {
         hex::encode(self.0)
     }
 
+    /// Provide the account identifier as bytes.
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
+
     /// Returns the checksum of the account identifier.
     pub fn generate_checksum(&self) -> [u8; 4] {
         let mut hasher = crc32fast::Hasher::new();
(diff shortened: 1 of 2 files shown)


library/ic-ledger-types/src/lib.rs
```diff
@@ -883,8 +888,11 @@ mod tests {
         )
     }
 
+    /// Verifies that these conversions yield the same result:
+    /// * bytes -> AccountIdentifier -> hex -> AccountIdentifier
+    /// * bytes -> AccountIdentifier
     #[test]
-    fn check_round_trip() {
+    fn check_hex_round_trip() {
```
bitdivine, line 895: Renamed to avoid ambiguity around hex/bytes round trip.

library/ic-ledger-types/src/lib.rs
```diff
@@ -218,6 +218,11 @@ impl AccountIdentifier {
         hex::encode(self.0)
     }
 
+    /// Provide the account identifier as bytes.
+    pub fn as_bytes(&self) -> &[u8; 32] {
+        &self.0
+    }
```
bitdivine, line 224: Note: The constructor takes great care to make sure that the checksum is valid, so this method is provided, rather than making the inner value public and compromising the integrity checks.

A fixed length array is returned rather than a slice, as otherwise the caller has no guarantee that the length is correct.

adamspofford-dfinity (approved):

PR #521, opened 2024-10-04
feat!: next generation of Rust CDK

lwshang:
SDK-1853

# Checklist:


src/ic-cdk/src/api/call.rs
```diff
@@ -189,6 +206,258 @@ unsafe extern "C" fn cleanup<T: AsRef<[u8]>>(state_ptr: *const RwLock<CallFuture
...
+    {
+        let fut = self.call_raw();
+        async {
+            let bytes = fut.await?;
+            decode_args(&bytes).map_err(decoder_error_to_reject::<R>)
+        }
+    }
+
+    /// Sends the call and decodes the reply to a Candid type with a decoding quota.
+    fn call_with_decoder_config<R: for<'b> ArgumentDecoder<'b>>(
```
chenyan-dfinity, line 361: With best-effort call, we can merge this function with the original `call` function. If the call is best-effort, put a decoding quota, if it's not, decode without the quota.

lwshang: The "decoding quota" and "best effort response" are orthogonal concepts. And they affect different stages of the inter-canister call. So I insist on keeping them separate.

sesi200 (approved):

PR #537, opened 2024-12-13
feat: Add bitcoin_get_block_headers

adamspofford-dfinity:
This intentionally calls the deprecated management canister function to be uniform with the rest; they will all be converted as a bloc.

### DIFF
--- a/src/ic-cdk/src/api/management_canister/bitcoin/types.rs
+++ b/src/ic-cdk/src/api/management_canister/bitcoin/types.rs
@@ -34,10 +34,10 @@ pub type BitcoinAddress = String;
 /// Block Hash.
 pub type BlockHash = Vec<u8>;
 
-/// Element in the Response of [bitcoin_get_current_fee_percentiles](super::bitcoin_get_current_fee_percentiles).
+/// Element in the Response of [`bitcoin_get_current_fee_percentiles`](super::bitcoin_get_current_fee_percentiles).
 pub type MillisatoshiPerByte = u64;
 
-/// Identifier of [Utxo].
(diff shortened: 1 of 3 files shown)


src/ic-cdk/src/api/management_canister/bitcoin/types.rs
```diff
@@ -34,10 +34,10 @@ pub type BitcoinAddress = String;
 /// Block Hash.
 pub type BlockHash = Vec<u8>;
 
-/// Element in the Response of [bitcoin_get_current_fee_percentiles](super::bitcoin_get_current_fee_percentiles).
+/// Element in the Response of [`bitcoin_get_current_fee_percentiles`](super::bitcoin_get_current_fee_percentiles).
 pub type MillisatoshiPerByte = u64;
 
-/// Identifier of [Utxo].
+/// Identifier of [`Utxo`].
```
lwshang, line 40: Were these backticks from auto-formatting?

I want to apply the non-backticks style in this project.

adamspofford-dfinity: No, I added them; there is a clippy::pedantic lint against code elements outside backticks.

Why put code elements in the prose font instead of the code font?

lwshang: I see. I had an impression that my code editor automatically removes the backticks in brackets. But that seems to be wrong.

I agree code font is better here.

I'll apply the new rule to the doc comments later.

lwshang (approved):

PR #548, opened 2025-01-16
feat: CallError redesign

lwshang:
# Description

There are a few reason we need a redesign of the CallError:
- The previous design made it hard to get the RejectCode which is the most common used field to do error handling.
- There is an initiative to expose error_code in system API. To avoid breaking changes at that time, we start to include error_code now.

## Error types

Aside from the `CallError`, I introduced `SystemError`. `SystemError` is a variant of `CallError`.

The error type of `call_raw()` and `call_oneway()` is `SystemError`. It implies that the call is rejected by the system.
```rust
pub struct SystemError {
    pub reject_code: RejectCode,
    pub reject_message: String,
    pub error_code: ErrorCode,
    pub sync: bool,
}
```
The error type of `call()` and `call_tuple` is `CallError`. They might also failed when decoding the reply.
```rust
pub enum CallError {
    CallRejected(SystemError),
    CandidDecodeFailed(String),
}
```

## Error codes

Borrowed the `ErrorCode` implementation from `pocket-ic`.

# How Has This Been Tested?

`ic-cdk-timers` internally make inter-canister calls to itself. The change there demonstrates that the error handling is simplified.

# Checklist:

### DIFF
--- a/ic-cdk-timers/src/lib.rs
+++ b/ic-cdk-timers/src/lib.rs
@@ -136,13 +136,8 @@ extern "C" fn global_timer() {
                 ic_cdk::println!("[ic-cdk-timers] canister_global_timer: {e:?}");
                 let mut retry_later = false;
                 match e {
-                    CallError::CallRejected(reject_code, _) => {
-                        if reject_code == RejectCode::SysTransient {
-                            retry_later = true;
-                        }
-                    }
-                    CallError::CallPerformFailed(reject_code, _) => {
-                        if reject_code == RejectCode::SysTransient {
+                    CallError::CallRejected(call_error) => {
+                        if call_error.reject_code == RejectCode::SysTransient {
                             retry_later = true;
                         }
                     }
--- a/ic-cdk-timers/src/lib.rs
+++ b/ic-cdk-timers/src/lib.rs
@@ -292,5 +292,5 @@ extern "C" fn timer_executor() {
             }
         }
     }
-    ic_cdk::api::msg_reply(&[]);
+    ic_cdk::api::msg_reply([]);
 }
--- a/ic-cdk-timers/src/lib.rs
+++ b/ic-cdk-timers/src/lib.rs
@@ -134,7 +134,7 @@ extern "C" fn global_timer() {
             let task_id = timer.task;
             if let Err(e) = res {
                 ic_cdk::println!("[ic-cdk-timers] canister_global_timer: {e:?}");
-                if e.reject_code == RejectCode::SysTransient {
+                if e.reject_code() == RejectCode::SysTransient {
                     // Try to execute the timer again later.
                     TIMERS.with(|timers| {
                         timers.borrow_mut().push(timer);
--- a/ic-cdk-timers/src/lib.rs
+++ b/ic-cdk-timers/src/lib.rs
@@ -29,7 +29,7 @@ use std::{
 use futures::{stream::FuturesUnordered, StreamExt};
 use slotmap::{new_key_type, KeyData, SlotMap};
 
-use ic_cdk::call::{Call, CallError, RejectCode, SendableCall};
+use ic_cdk::call::{Call, RejectCode, SendableCall};
 
 // To ensure that tasks are removable seamlessly, there are two separate concepts here: tasks, for the actual function being called,
 // and timers, the scheduled execution of tasks. As this is an implementation detail, this does not affect the exported name TimerId,
@@ -117,8 +117,8 @@ extern "C" fn global_timer() {
                                         ic_cdk::api::canister_self(),
                                         "<ic-cdk internal> timer_executor",
                                     )
-                                    .with_arg(task_id.0.as_ffi())
-                                    .call::<()>()
+                                    .with_raw_args(task_id.0.as_ffi().to_be_bytes().to_vec())
+                                    .call_raw()
                                     .await,
                                 )
                             });
@@ -135,15 +135,8 @@ extern "C" fn global_timer() {
             if let Err(e) = res {
                 ic_cdk::println!("[ic-cdk-timers] canister_global_timer: {e:?}");
                 let mut retry_later = false;
-                match e {
-                    CallError::CallRejected(call_error) => {
-                        if call_error.reject_code == RejectCode::SysTransient {
-                            retry_later = true;
-                        }
-                    }
-                    CallError::CandidDecodeFailed(_) => {
-                        // This error is not transient, and will not be retried.
-                    }
+                if e.reject_code == RejectCode::SysTransient {
+                    retry_later = true;
                 }
                 if retry_later {
                     // Try to execute the timer again later.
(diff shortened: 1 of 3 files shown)


ic-cdk-timers/src/lib.rs
```diff
@@ -135,20 +135,8 @@ extern "C" fn global_timer() {
             if let Err(e) = res {
                 ic_cdk::println!("[ic-cdk-timers] canister_global_timer: {e:?}");
                 let mut retry_later = false;
```
ericswanson-dfinity, line 137:
```suggestion
                let retry_later = e.reject_code == RejectCode::SysTransient;
```

ic-cdk/src/call.rs
```diff
@@ -163,7 +281,35 @@ pub enum CallError {
...
+    /// When the call was rejected synchronously (`ic0.call_preform` returns non-zero code),
+    /// this message is set to a fixed string ("failed to enqueue the call").
+    pub reject_message: String,
+    /// See [`ErrorCode`].
+    ///
+    /// # Note
+    ///
+    /// As of the current version of the IC, the error codes are not available in the system API.
+    /// Please DO NOT rely on the error codes until they are officially supported.
+    pub error_code: ErrorCode,
```
ericswanson-dfinity, line 303: What's the purpose of having this 'pub' when the value is generally incorrect (from reject_to_error) and the error codes haven't been officially finalized?

lwshang: We can remove the pub for now. But when ic0.msg_error_code is officially supported, we will need to add back the pub. Then that will be a breaking change.

One alternative is to make the fields private and provide public getters. But that will harm the UX when users want to extract multiple fields.

After more thinking, I feel that the alternative approach above could be better. I will followup with Martin.

ericswanson-dfinity (approved):

ic-cdk-timers/src/lib.rs
```diff
@@ -117,8 +117,8 @@ extern "C" fn global_timer() {
                                         ic_cdk::api::canister_self(),
                                         "<ic-cdk internal> timer_executor",
                                     )
-                                    .with_arg(task_id.0.as_ffi())
-                                    .call::<()>()
+                                    .with_raw_args(task_id.0.as_ffi().to_be_bytes().to_vec())
+                                    .call_raw()
```
mraszyk, line 121: Why did you switch to "raw" here?

ic-cdk/src/call.rs
```diff
@@ -31,41 +28,47 @@ pub enum RejectCode {
...
-    fn from(code: u32) -> Self {
+impl Error for InvalidRejectCode {}
+
+impl TryFrom<u32> for RejectCode {
+    type Error = InvalidRejectCode;
+
+    fn try_from(code: u32) -> Result<Self, Self::Error> {
         match code {
-            // 0 is a special code meaning "no error"
-            0 => RejectCode::NoError,
```
mraszyk, line 46: would it make sense to assert that the code is not zero?

ic-cdk/src/call.rs
```diff
@@ -163,7 +98,46 @@ pub enum CallError {
...
+impl CallRejected {
+    /// Returns the [`RejectCode`].
+    pub fn reject_code(&self) -> RejectCode {
+        self.reject_code
+    }
+
+    /// Returns the reject message.
+    ///
+    /// When the call was rejected asynchronously (IC rejects the call after it was enqueued),
+    /// this message is set with [`msg_reject`](crate::api::msg_reject).
```
mraszyk, line 122: the reject message can also be produced by the system (canister stopped, out of cycles etc.) or come from crate::api::trap (including a wrapper by the system in case of traps)

ic-cdk/src/call.rs
```diff
@@ -163,7 +98,46 @@ pub enum CallError {
...
+    ///
+    /// When the call was rejected synchronously (`ic0.call_preform` returns non-zero code),
+    /// this message is set to a fixed string ("failed to enqueue the call").
```
mraszyk, line 125: I'd recommend to stick to the current message:
- for the sake of backward-compatibility (no change in behavior)
- "failed to enqueue the call" sounds like the queue is the issue, but the failure could also be due to cycles (for instance)

ic-cdk/src/call.rs
```diff
@@ -397,8 +381,11 @@ impl SendableCall for Call<'_> {
 }
 
 impl<'a, T: ArgumentEncoder + Send + Sync> SendableCall for CallWithArgs<'a, T> {
-    async fn call_raw(self) -> CallResult<Vec<u8>> {
-        let args_raw = encode_args(self.args).map_err(encoder_error_to_call_error::<T>)?;
+    async fn call_raw(self) -> SystemResult<Vec<u8>> {
+        // Candid Encoding can only fail if heap memory is exhausted.
```
mraszyk, line 385: if the heap memory is exhausted, the canister traps anyway

ic-cdk/src/call.rs
```diff
@@ -536,11 +532,17 @@ impl<T: AsRef<[u8]>> Future for CallFuture<T> {
...
-                        let result = Err(CallError::CallPerformFailed(c));
+                match code {
+                    0 => {
+                        // call_perform returns 0 means the call was successfully enqueued.
+                    }
+                    _ => {
+                        let reject_code = RejectCode::try_from(code).unwrap();
+                        let result = Err(CallRejected {
+                            reject_code,
+                            reject_message: "failed to enqueue the call".to_string(),
```
mraszyk, line 543: Could we define this as a string constant given that it is used at multiple places?

ic-cdk-timers/src/lib.rs
lwshang: To avoid unnecessary Candid encoding.

ic-cdk/src/call.rs
lwshang: There is no need of assertion right? Since there is no variant for 0 now. And the RejectCode instances are constructed using `TryFrom<u32>` which will panic for 0.

ic-cdk-timers/src/lib.rs
mraszyk: That makes sense!

ic-cdk/src/call.rs
```diff
@@ -565,9 +567,16 @@ unsafe extern "C" fn callback<T: AsRef<[u8]>>(state_ptr: *const RwLock<CallFutur
     if let Some(state) = state.upgrade() {
         // Make sure to un-borrow_mut the state.
         {
-            state.write().unwrap().result = Some(match RejectCode::from(msg_reject_code()) {
-                RejectCode::NoError => Ok(msg_arg_data()),
-                c => Err(CallError::CallRejected(c, msg_reject_msg())),
+            state.write().unwrap().result = Some(match msg_reject_code() {
+                0 => Ok(msg_arg_data()),
+                code => {
+                    let reject_code = RejectCode::try_from(code).unwrap();
```
mraszyk, line 573: this unwrap could be dangerous for canisters with an outdated CDK version if the replica introduces a new reject code

ic-cdk/src/call.rs
lwshang: That's a typo.
I wanted to mention 
```
[`msg_reject_msg`](crate::api::msg_reject_msg).
```

ic-cdk/src/call.rs
mraszyk: > And the RejectCode instances are constructed using TryFrom<u32> which will panic for 0.

I missed that panic which is actually dangerous - please see my other comment. Panicking in case of 0 makes sense though since 0 can never become a new reject code.

PR #563, opened 2025-02-27
feat: Add types for fetching canister logs

michael-weigelt:
# Description

The specification mentions these types, but they are not in the ic-management-canister-types yet. 

# How Has This Been Tested?

The new type was added to the candid_equality.rs test

# Checklist:

### DIFF
--- a/ic-management-canister-types/CHANGELOG.md
+++ b/ic-management-canister-types/CHANGELOG.md
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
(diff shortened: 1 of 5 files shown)


lwshang:
The CI failures are caused by the indirect dependency crates (`litemap`, `zeromap` and more `icu*`) from the `icu4x` project. Yesterday, they cut releases in which the MSRV was bumped but the major versions were not bumped accordingly.

It seems that, these indirect dependencies are introduced by `pocket-ic` in `e2e-tests`. In another PR, I will soon separate the `e2e-tests` crate out from the repo workspace so that it can use a higher version of Rust toolchain than the lib crates.

PR #567, opened 2025-03-03
fix: only call `call_cycles_add128` if cycles amount > 0

hpeebles:
# Description

Skip calling `ic0::call_cycles_add128` if the cycles amount is 0. In OpenChat all of our c2c calls fall into the same underlying method where we pass in the number of cycles to add. For almost all requests this is 0, but `call_cycles_add128` is still being called which causes composite queries to fail.

# How Has This Been Tested?

The OpenChat tests now pass with this change included, whereas previously the composite queries were failing - https://github.com/open-chat-labs/open-chat/pull/7506

# Checklist:


lwshang:
Currently, the cycles field is an `Option<u128>`, and the `ic0::call_cycles_add` API is only invoked when `cycles` is Some. This makes sense because:

- If `with_cycles()` is not called, the field remains `None`, and the API is not invoked.
- If `with_cycles(0)` is called, the field becomes `Some(0)`, and the API is invoked with 0.

While invoking the API with 0 might seem unnecessary, it’s a rare edge case because `with_cycles(0)` is logically meaningless in most real-world applications. Over-engineering for this edge case adds unnecessary complexity without significant benefit.

I suggest we keep the current behavior for simplicity and clarity.

hpeebles:
Fair enough.

But the downside as a consumer of this CDK is that the code becomes a bit messy if all calls flow through the same path (which is what we do in OpenChat for all c2c calls).

See here - https://github.com/open-chat-labs/open-chat/pull/7506/commits/d687921a2ddab2872ddbe4178a09d15170a3c7ea

I have had to replace 
```
let response = ic_cdk::call::Call::unbounded_wait(canister_id, method_name)
        .with_raw_args(payload_bytes)
        .with_cycles(cycles)
        .await;
```

With
```
let mut call = ic_cdk::call::Call::unbounded_wait(canister_id, method_name).with_raw_args(payload_bytes);

if cycles > 0 {
    call = call.with_cycles(cycles);
}

let response = call.await;
```

It's not the end of the world, but it's definitely not as nice as it could be.
I guess one obvious counter argument would be that we should have 1 function for calling with cycles, and one for calling without, which we definitely could do, but it just seems like more code unnecessarily.

hpeebles:
See what you think of this version, if you are happy with it I can turn it into a PR - https://github.com/hpeebles/cdk-rs/commit/e8a116fceb7bff23e5440e01e592c32866c66360

I've made the `cycles` field be a `u128` rather than an `Option<u128>`, so the check becomes if  > 0 rather than if `Some(..)`.

lwshang:
> I've made the cycles field be a `u128` rather than an `Option<u128>`

Good idea. I reopened this PR. Please update your work here. I'll merge it soon.

hpeebles:
Just to confirm, the OpenChat tests all pass as of this commit - https://github.com/open-chat-labs/open-chat/pull/7506/commits/a80ef1bbbbe2171decf9c291696d8e0b42e74808

PR #570, opened 2025-03-05
feat: add cost_* System API and methods that calculate cycles cost under the hood

lwshang:
SDK-1963

# Description

- Updated `ic0` according to https://github.com/dfinity/portal/pull/4110.
- Added bindings for `cost_*` System API in `api.rs`.
- Removed the `_with_cycles` suffix from some Management canister methods
  - They no longer takes `cycles` as an argument.
  - The cycles cost is calculated inside by invoking the new `cost_*` API.
 
Note: No binding for `cost_vetkd_derive_encrypted_key` because the corresponding Management canister methods hasn't been included.

# How Has This Been Tested?

Updated related e2e tests.

# Checklist:

- [x] The title of this PR complies with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- [x] I have edited the CHANGELOG accordingly.
- [x] I have made corresponding changes to the documentation.

PR #574, opened 2025-03-10
feat: Add higher level cost APIs 

michael-weigelt:
To be merged or cherry-picked into https://github.com/dfinity/cdk-rs/pull/570

### DIFF
--- a/ic-cdk/src/call.rs
+++ b/ic-cdk/src/call.rs
@@ -41,7 +41,7 @@
 //! The module also includes internal types and functions to manage the state and execution of inter-canister calls,
 //! such as [`CallFuture`] and its associated state management.
 
-use crate::api::{msg_arg_data, msg_reject_code, msg_reject_msg};
+use crate::api::{cost_call, msg_arg_data, msg_reject_code, msg_reject_msg};
 use crate::{futures::is_recovering_from_trap, trap};
 use candid::utils::{encode_args_ref, ArgumentDecoder, ArgumentEncoder};
 use candid::{decode_args, decode_one, encode_one, CandidType, Deserialize, Principal};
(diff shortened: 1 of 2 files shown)


ic-cdk/src/call.rs
```diff
@@ -266,6 +266,20 @@ impl<'a> Call<'_, 'a> {
...
+    /// - the payload length
+    /// - the cost of transmitting the request
+    /// - the cost for the reservation of response transmission (may be partially refunded)
+    /// - the cost for the reservation of callback execution (may be partially refunded).
+    pub fn get_cost(&self) -> u128 {
+        self.cycles.saturating_add(cost_call(
+            self.method.len() as u64,
+            self.encoded_args.len() as u64,
+        ))
+    }
```
michael-weigelt, line 282: @mraszyk did I forget some call cost factor here?

ic-cdk/src/management_canister.rs
```diff
@@ -610,6 +623,16 @@ pub async fn schnorr_public_key(arg: &SchnorrPublicKeyArgs) -> CallResult<Schnor
     )
 }
 
+/// Calculates the cost of signing with the given Schnorr argument.
+/// See [`sign_with_ecdsa`] for more details.
+pub fn cost_sign_with_schnorr(arg: &SignWithSchnorrArgs) -> Result<u128, SignCostError> {
+    let algorithm = match arg.key_id.algorithm {
+        SchnorrAlgorithm::Bip340secp256k1 => 0,
+        SchnorrAlgorithm::Ed25519 => 1,
```
michael-weigelt, line 631: What do you think about associating this mapping with the type directly, e.g. as `impl Into<u32>`?

ic-cdk/src/call.rs
mraszyk: I think so: https://github.com/dfinity/ic/blob/cc7b6f155b9334ca1406435bc7bbcaf1d547f9e1/rs/types/types/src/messages/inter_canister.rs#L136

ic-cdk/src/management_canister.rs
mraszyk: I think that's a good idea.

lwshang (approved):

PR #603, opened 2025-05-13
chore: Factor out v2 executor into separate crate

adamspofford-dfinity:
This PR targets a feature branch because it is for ic-cdk@0.17, not 0.18.

### DIFF
--- a/src/ic-cdk-executor/Cargo.toml
+++ b/src/ic-cdk-executor/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "ic-cdk-executor"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+license.workspace = true
+description = "Async executor for `ic-cdk`"
+categories = ["asynchronous", "rust-patterns"]
+keywords = ["internet-computer", "dfinity", "canister", "cdk"]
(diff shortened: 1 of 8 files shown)


src/ic-cdk-executor/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+edition.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+license.workspace = true
+description = "Async executor for `ic-cdk`"
+categories = ["asynchronous", "rust-patterns"]
+keywords = ["internet-computer", "dfinity", "canister", "cdk"]
+# DO NOT change this field, not even to update the link, unless you are updating every existing version of the package.
+# Update the link by updating the links-pin tag.
+links = "ic-cdk async executor, see https://github.com/dfinity/cdk-rs/blob/links-pin/TROUBLESHOOTING.md"
```
lwshang, line 14: What will this field be in `ic-cdk-executor` v0.2.0 for `ic-cdk` v0.18?
And what will be in the `build.rs` script there?

adamspofford-dfinity: This field will always stay the same and nothing will be in the build.rs script. Cargo emits an error if you have `links` without `build.rs`, so you have to have `build.rs`, but there's nothing in it.

lwshang (approved):

PR #624, opened 2025-06-11
feat: Add safe interface to ic0

adamspofford-dfinity:


ic0/src/lib.rs
```diff
@@ -8,5 +8,594 @@
     clippy::missing_safety_doc
 )]
 
-mod ic0;
-pub use crate::ic0::*;
+use std::mem::MaybeUninit;
```
lwshang, line 11: Could you provide a module level doc to summarize what methods are expected in this module?
"safe Rust translation of system API" is not an accurate/complete statement:
- `call_new` and `call_on_cleanup` are still `unsafe` functions.
- Several methods have a `*_uninit` twin, we should better have an explanation.

ic-cdk/src/api/call.rs
```diff
@@ -110,40 +110,49 @@ impl<T: AsRef<[u8]>> Future for CallFuture<T> {
             } => {
                 let callee = id.as_slice();
                 let args = arg.as_ref();
-                let state_ptr = Arc::into_raw(Arc::clone(&self_ref.state));
+                let state_ptr =
+                    Arc::<RwLock<CallFutureState<T>>>::into_raw(Arc::clone(&self_ref.state));
```
lwshang, line 114: The type is written more explicitly. I saw the `callback` and `cleanup` functions' comment also mentions this type. Is this intended to check that `self_ref.state` has the correct type?

lwshang (commented):
LGTM. Just a few requests for clarification or more documentation.

IIUC, this is a breaking change to `ic0` crate itself, but not a breaking change to the dependent crates like `ic-cdk`, right?

ic0/README.md
```diff
@@ -4,11 +4,11 @@ Internet Computer System API binding.
...
 
-`ic0` is simply an unsafe Rust translation of the System API as described in the [IC interface specification][1].
+`ic0` is simply a safe Rust translation of the System API as described in the [IC interface specification][1]. The unsafe direct imports can be found in the `ic0::sys` module.
 
 ## Update
 
-`ic0` keeps in step with the IC interface specification. Particularly, `ic0` is directly generated from [system API][1] in that repo.
+`ic0` keeps in step with the IC interface specification. Particularly, `ic0` is directly generated from the [system API][1] in that repo.
 
 When interface-spec releases a new version that modify [system API][1]:
```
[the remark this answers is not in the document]
adamspofford-dfinity: Sounds like a good idea

adamspofford-dfinity:
Yes, this is breaking for ic0.

ic0/Cargo.toml
```diff
@@ -11,6 +11,7 @@ readme = "README.md"
 categories = ["api-bindings", "development-tools::ffi"]
 keywords = ["internet-computer", "types", "dfinity", "canister", "cdk"]
 include = ["src", "Cargo.toml", "LICENSE", "README.md"]
+links = "ic0"
```
lwshang, line 14: Why add links here?
IIUC, multiple versions of `ic0` can exist in a dependency tree since it doesn't set global state nor define exports.

adamspofford-dfinity: ah, I think I misunderstood its real purpose - you are correct

PR #625, opened 2025-06-16
feat: convert timers to async

adamspofford-dfinity:
In case it is not clear, the additional trait is because AsyncFnMut is not dyn-compatible, while the choice of Future vs AsyncFnOnce is because there is no functional difference between the two when there are no parameters, so the extra internal machinery to AsyncFnOnce is wasted.


ic-cdk-timers/src/lib.rs
```diff
@@ -281,15 +293,32 @@ extern "C" fn timer_executor() {
...
+                        guard.0.as_mut().unwrap().call_mut().await;
+                        impl Drop for RepeatGuard {
+                            fn drop(&mut self) {
+                                TASKS.with(|tasks| {
+                                    tasks.borrow_mut().get_mut(self.1).map(|slot| {
+                                        *slot = Task::Repeated {
+                                            func: self.0.take().unwrap(),
+                                            interval: self.2,
+                                        }
+                                    })
```
gregorydemay, line 316: nit: shouldn't the closure variable `slot` be named `task` (or something similar). If I understood correctly, we are resetting the task at the given `TimerId`.

ic-cdk-timers/src/lib.rs
```diff
@@ -281,15 +293,32 @@ extern "C" fn timer_executor() {
         let mut tasks = tasks.borrow_mut();
         tasks.get_mut(task_id).map(mem::take)
     });
-    if let Some(mut task) = task {
+    if let Some(task) = task {
         match task {
-            Task::Once(func) => {
-                ic_cdk::futures::in_executor_context(func);
+            Task::Once(fut) => {
+                ic_cdk::futures::in_executor_context(|| ic_cdk::futures::spawn(fut));
```
gregorydemay, line 299: Maybe it should be worth highlighting (PR description or release notes/changelog) that although it uses the new `spawn` (from 0.18) in that case it doesn't change the behavior of the executed code (if I remember the discussion on notion correctly)

gregorydemay (approved):
Some minor comments/understanding questions from my side but otherwise LGTM!

ic-cdk-timers/src/lib.rs
adamspofford-dfinity: It represents a storage slot, which for a little bit here, *doesn't* contain a task. self.1 is the real task.

ic-cdk-timers/src/lib.rs
```diff
@@ -43,15 +45,25 @@ thread_local! {
 
 enum Task {
     Repeated {
-        func: Box<dyn FnMut()>,
+        func: Box<dyn RepeatedClosure>,
```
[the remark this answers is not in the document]
gregorydemay: Thanks for the explanation!

ic-cdk-timers/src/lib.rs
gregorydemay: I see, that's a fait point. Thanks for the explanation.

ic-cdk-timers/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! ```rust,no_run
 //! # use std::time::Duration;
 //! # fn main() {
-//! ic_cdk_timers::set_timer(Duration::from_secs(1), || ic_cdk::println!("Hello from the future!"));
+//! ic_cdk_timers::set_timer(Duration::from_secs(1), async { ic_cdk::println!("Hello from the future!") });
```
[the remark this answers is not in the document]
gregorydemay: No problem! Would that be part of this PR?

PR #641, opened 2025-07-24
feat: remove ic-cdk dependency from ic-cdk-timers

adamspofford-dfinity:


ic-cdk-timers/Cargo.toml
```diff
@@ -10,23 +10,19 @@ description = "Timers library for the Rust CDK."
...
-    "no-std",
-    "development-tools::ffi",
-]
+categories = ["api-bindings", "data-structures", "no-std", "development-tools::ffi"]
 keywords = ["internet-computer", "dfinity", "canister", "cdk"]
 include = ["src", "Cargo.toml", "LICENSE", "README.md"]
 
 [dependencies]
-candid.workspace = true
+futures.workspace = true
```
lwshang, line 18: It seems that we no longer depends on `futures`.
Should be safe to remove it here and in the root `Cargo.toml`.

lwshang (commented):
LGTM. Ready for the security review.

I left a few minor suggestions for cleanup.

ic-cdk-timers/Cargo.toml
adamspofford-dfinity: It's used in e2e-tests, so not the root.

ic-cdk-timers/src/lib.rs
```diff
@@ -102,140 +101,240 @@ impl Eq for Timer {}
...
-                                        });
-                                    }),
-                                    None => ic_cdk::println!(
-                                        "Failed to reschedule task (needed {interval}, currently {now}, and this would exceed u64::MAX)",
-                                        interval = interval.as_nanos(),
-                                    ),
-                                }
+                            ic0::call_with_best_effort_response(300);
+                            ic0::call_data_append(task_id.0.as_ffi().to_be_bytes().as_ref());
+                            let errcode = ic0::call_perform();
```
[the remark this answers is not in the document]
adamspofford-dfinity: The original used FuturesUnordered - doesn't that mean it's not breaking?

lwshang (approved):
LGTM.
It'll be better to add some examples for the public methods.
Also, this is a breaking change for `ic-cdk-timers`. Let's make it more clear in the PR title and the changelog.

PR #646, opened 2025-08-05
chore: remove macos tests

adamspofford-dfinity:
Our tests are all wasm-targeted, with x86 code only being used for harnesses. We do not need to use expensive macos machines for them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,11 +15,7 @@ concurrency:
 jobs:
   msrv:
     name: cargo build with MSRV
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


lwshang:
I recommend we keep the macOS tests.

While the CDK target Wasm, the compilation process and e2e testing are dependent on the host OS. The macOS tests are essential for ensuring that:

- The build.rs script and other build logic work correctly on macOS.
- Users can successfully develop Canister Wasm modules on both macOS and Linux.
- The OS-specific PocketIC server binaries function as expected.

Removing these tests could lead to bugs that only affect our macOS users. The cost is justified to ensure cross-platform compatibility.

raymondk (approved):

PR #665, opened 2025-09-19
feat(mgmt-types)!: Add new fields to `CanisterStatus` and `SubnetInfo`

michael-weigelt:
This PR adds new fields to management endpoints which are needed for canister migration.


lwshang:
Done in #677

PR #708, opened 2026-04-16
feat: ic0 v1.1.0 and ic-cdk v0.20.1 with msg_caller_info_{data,signer} APIs

lwshang:
## Summary

- Add `msg_caller_info_data_size/copy` and `msg_caller_info_signer_size/copy` raw bindings to `ic0` (sys.rs and lib.rs)
- Add `msg_caller_info_data() -> Vec<u8>` and `msg_caller_info_signer() -> Option<Principal>` higher-level bindings to `ic_cdk::api`
- Add e2e tests using PocketIC's `update_call_with_sender_info`

These APIs expose the caller identity attribute feature described in [dfinity/portal#6202](https://github.com/dfinity/portal/pull/6202). `msg_caller_info_data` returns auxiliary data provided by the signing canister, and `msg_caller_info_signer` returns the canister ID of the canister that provided the caller's canister signature. Both return empty/`None` when the caller is not authenticated via canister signatures.

## PocketIC setup improvements

The e2e test harness/CI now supports pinning `pocket-ic` to a git revision in addition to a tag, and the download step resolves the matching server binary for either form. This unblocked developing against an unreleased RC of PocketIC before cutting over to the tagged release.

## Crate releases

- `ic0` 1.0.1 → 1.1.0
- `ic-cdk` 0.20.0 → 0.20.1 (and `ic-cdk-macros` bumped in lockstep)

## Test plan

- [x] E2e tests pass: `call_msg_caller_info_data` and `call_msg_caller_info_signer` verify correct values with and without `sender_info`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/e2e-tests/tests/api.rs
+++ b/e2e-tests/tests/api.rs
@@ -8,8 +8,8 @@ use test_utilities::{cargo_build_canister, pic_base, update};
 #[test]
 fn call_api() {
     let wasm = cargo_build_canister("api");
-    // with_ii_subnet is required for testing the ic0.cost_sign_with_* API with pre-defined key name.
-    let pic = pic_base().with_ii_subnet().build();
+    // with_test_threshold_keys_subnet is required for testing the ic0.cost_sign_with_* API with pre-defined key name.
+    let pic = pic_base().with_test_threshold_keys_subnet().build();
     let canister_id = pic.create_canister();
     pic.add_cycles(canister_id, 100_000_000_000_000);
     pic.install_canister(canister_id, wasm, vec![], None);
--- a/e2e-tests/tests/api.rs
+++ b/e2e-tests/tests/api.rs
@@ -1,6 +1,6 @@
 use candid::Principal;
 use ic_cdk_management_canister::{CanisterSettings, EnvironmentVariable, UpdateSettingsArgs};
-use pocket_ic::ErrorCode;
+use pocket_ic::{ErrorCode, common::rest::RawSenderInfo};
 
 mod test_utilities;
 use test_utilities::{cargo_build_canister, pic_base, update};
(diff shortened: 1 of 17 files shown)


e2e-tests/tests/api.rs
```diff
@@ -25,6 +25,43 @@ fn call_api() {
...
+    let res = pic
+        .update_call(canister_id, sender, "call_msg_caller_info_data", vec![])
+        .unwrap();
+    assert!(res.is_empty());
+    // With sender_info: signer should be the canister_id principal bytes.
+    let res = pic
+        .update_call_with_sender_info(
+            canister_id,
+            sender,
+            "call_msg_caller_info_signer",
```
aterga, line 54: Seems to be a bit inconsistent between 

`call_msg_caller_info_signer`
`sender_info`

Therefore, consider changing the convention to:

```suggestion
            "call_msg_sender_info_signer",
```

lwshang: I don't think there is inconsistency.

  - System API: `msg_caller_info_signer_size` / `msg_caller_info_signer_copy`
  - High-level binding (ic-cdk/src/api.rs): `msg_caller_info_signer`
  - E2E test canister endpoint (e2e-tests/src/bin/api.rs): `call_msg_caller_info_signer` — simply invokes the `msg_caller_info_signer` binding.

`sender_info`, on the other hand, is the field name in the HTTP ingress message and the corresponding argument to pocket-ic's `update_call_with_sender_info`.

aterga: Okay, thanks for checking.

e2e-tests/tests/api.rs
```diff
@@ -25,6 +25,43 @@ fn call_api() {
     // Unlike the other entry points, `call_msg_dealine_caller` was implemented with the `#[update]` macro.
     // So we use the update method which assumes candid
     let _: () = update(&pic, canister_id, "call_msg_deadline_caller", ()).unwrap();
+    let info_data = b"test_caller_info_data";
+    let sender_info = RawSenderInfo {
+        info: info_data.to_vec(),
+        signer: canister_id.as_slice().to_vec(),
+    };
+    // With sender_info: data should equal the info bytes that were sent.
```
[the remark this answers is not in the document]
lwshang: The test is specifically verifying API binding mechanics: that `msg_caller_info_data` and `msg_caller_info_signer` return the values from `sender_info`, not from `sender`. The assertions check against `sender_info.signer` (which is `canister_id.as_slice()`), independent of the sender's principal type. PocketIC passes `sender_info` through regardless of the sender type, so using `Principal::anonymous()` correctly exercises the binding without adding unnecessary complexity.

scripts/download_pocket_ic_server.sh
```diff
@@ -2,23 +2,54 @@
...
 cd "$SCRIPTS_DIR/.."
-# extract the tag from the workspace Cargo.toml
-tag=$(grep -E 'pocket-ic.*tag' Cargo.toml | sed -n "s/.*tag *= *\"\([^\"]*\)\".*/\1/p")
+
+# Extract the source URL of the pocket-ic dependency via cargo metadata.
+# With --no-deps only workspace packages are emitted; .dependencies[] still
+# carries the declared source (e.g. git+https://...?rev=<hash>#<hash>).
+pocket_ic_source=$(cargo metadata --format-version=1 --no-deps \
+    | jq -r '.packages[].dependencies[] | select(.name=="pocket-ic") | .source' \
+    | head -1)
```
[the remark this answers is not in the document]
lwshang: This script is intended for repo maintainers and CI only, where both `cargo` and `jq` are always present. Defensive checks for these tools would add noise without benefit in this context.

adamspofford-dfinity (approved):

